PR #30, opened 2018-10-23
encoding-fix

Vice93:
Encoding on a few documents are wrong.


Vice93:
Ugh, I want to push a branch to fix it, but I haven't done this before...

PR #33, opened 2018-11-05
Removed colon from weapon_category:

Arbust908:
+Unneeded colon on  the weapon_category: on 5e-SRD-Equipment.json


Tenari:
yeah I did the same thing and was about to pr this

PR #56, opened 2019-04-11
Remove duplicated entries from Equipment table, indices 226-250

syntaxaire:
Unclear how these were entered, since in some cases they are exact
duplicates, but in other cases the weight of the duplicate item is 1/10
that of the entry found in the SRD.


bagelbits:
That's super weird.

PR #67, opened 2019-05-20
Fix typical speakers based on the SRD:

ogregoire:
* change deep speech typical speakers from mindflayers and beholders to aboleths and cloakers;
* remove tieflings from infernal typical speakers;
* fix word casing for fey creatures and underdark traders.

See SRD 5.1, page 59.


bagelbits:
Looks good!

PR #69, opened 2019-05-23
Remove class descriptions not part of the SRD

ogregoire:
They actually come from the PHB and are not covered by OGL.


bagelbits:
Thanks! Feel free to flag anything else like this. I know there is some hiding everywhere.

bagelbits (approved):

Issue #68, opened 2019-05-21
Can some values be null?


ogregoire:
Yesterday I wanted to clean up the language file, which I did. But I noticed that in the SRD, a big line strikes through the script column of the "deep speech" language. This basically means that there is no script.

However the file https://github.com/adrpadua/5e-database/blob/master/5e-SRD-Languages.json currently contains the actual value "none". There is no indicator that there is no script: the script currently has the value "none". In my view, this is incorrect.

There are usually two alternatives to this:

* the first is to remove the field entirely. If we do that, then the file is inconsistent with other files because the key is usually present but maps to an empty value, usually an array. Check the race file: it has that behavior.
* the second alternative is to use `null`. This way all the fields are set, but the actual value is set to null, and not "none". I personally like it less than the first alternative, but it keeps the consistency with the other files and reduce the work to provide.

So in short, we have three options:

1. Keep as is, having the field set to a dummy value

        {
        	"index": 12,
        	"name": "Deep Speech",
        	"type": "Exotic",
        	"typical_speakers": ["Aboleths", "Cloakers"],
        	"script": "none",
        	"url": "http://www.dnd5eapi.co/api/languages/12"
        }

2. Remove the field, opening the door to inconsistency.

        {
        	"index": 12,
        	"name": "Deep Speech",
        	"type": "Exotic",
        	"typical_speakers": ["Aboleths", "Cloakers"],
        	"url": "http://www.dnd5eapi.co/api/languages/12"
        }

3. Change `"none"` to `null` so that the key is present, but the value effectively shows that there is no value.

        {
        	"index": 12,
        	"name": "Deep Speech",
        	"type": "Exotic",
        	"typical_speakers": ["Aboleths", "Cloakers"],
        	"script": null,
        	"url": "http://www.dnd5eapi.co/api/languages/12"
        }

In the future, it might be good to tend towards 2 for everything, but it might require too much work at once. So I strongly suggest that `null` is accepted for the time being.

bagelbits:
Technically four options if you count:
```json
 {
 	"index": 12,
 	"name": "Deep Speech",
 	"type": "Exotic",
 	"typical_speakers": ["Aboleths", "Cloakers"],
 	"script": "",
 	"url": "http://www.dnd5eapi.co/api/languages/12"
 }
```

I'd honestly be fine with either that or `null`. Though, it looks like we're not consistent on either approach. It might be worth considering?

bagelbits:
[...]

PR #73, opened 2019-06-04
Clean up none and empty string usecases #68

bagelbits:
Some cleanup for issue #68


bagelbits:
@ogregoire What do you think?

ogregoire:
This seems nice. Cleaning up is done incrementally and this is an excellent start!

PR #74, opened 2019-06-05
Cleaning up substype and spell

bagelbits:
Just some more cleanup.


bagelbits:
@ogregoire Some more clean up for #68

ogregoire:
This seems good!

PR #75, opened 2019-06-06
Clean up monster attrs

bagelbits:
More cleanup based on my conversations with @ogregoire in #68 
This is converting `damage_vulnerabilities`, `damage_resistances`, `damage_immunities`, and `condition_immunities` to be lists instead of strings while still preserving all data.


bagelbits:
@ogregoire More cleanup for #68 
It's a lot.

ogregoire:
@bagelbits that's a lot of work! Very nice 😀 I'm currently on holiday since already a week hence the fact that I didn't do this myself, but you're doing a very good job with making the API much more data oriented than it is.

bagelbits:
Haha thanks! I'll throw up another PR with the senses and languages for monsters. However, I'll let you own `spell_acquisition_method` changes.

PR #77, opened 2019-06-11
Merge spell acquisition method into prerequisites

ogregoire:


bagelbits (approved):
Awesome! That's much cleaner and makes more sense.

PR #79, opened 2019-08-08
Clarify license detail

tyndyll:
Moved the `mit.md` file to the more standard LICENSE.md naming. Called out the license in the README and made reference to the OGL license


bagelbits:
That looks great! Sorry. I was on vacation for a week.

PR #80, opened 2019-08-08
fix level requirement for fighter Action Surge

Tenari:


bagelbits:
That looks great! Sorry. I was on vacation.

PR #90, opened 2019-10-01
Race Trait Url correction, added missing traits, added missing subraces

Incognitonomous:
Mostly just corrected the urls, but also added some missing traits and added two subraces missing from the elven race (Woodland Elf and Dark Elf)

### DIFF
--- a/5e-SRD-Races.json
+++ b/5e-SRD-Races.json
@@ -75,8 +75,12 @@
 				"url": "http://www.dnd5eapi.co/api/traits/2"
 			},
 			{
-				"name": "Stonecunning",
+				"name": "Dwarven Combat Training",
 				"url": "http://www.dnd5eapi.co/api/traits/3"
+			},
+			{
+				"name": "Stonecunning",
+				"url": "http://www.dnd5eapi.co/api/traits/4"
 			}
 		],
 		"subraces": [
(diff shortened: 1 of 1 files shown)


ogregoire:
Woodland elf and dark elf are not part of the SRD, but are part of the PHB, This means that they are not covered by the OGL which permits their reuse such as in this project.

Same for the "missing traits": they're not part of the SRD, but they're indeed in the PHB.

bagelbits:
@ogregoire is correct on this. @Incognitonomous, I'm happy to take the rest of the changes if you drop the added subraces that aren't apart of OGL.

ogregoire:
The trait URL changes are consistent with the API as available on the website, which doesn't represent the latest changes in this repo. I wouldn't accept that change because, as of now, the race trait URLs are consistent with the traits in the traits JSON file.

In short, it would be a regression.

bagelbits:
@ogregoire Thanks for pointing that out! I'll close this out then.

Incognitonomous:
In what way would it be a regression? when using the api, i'm recieving the incorrect url for the corresponding traits, not at all consistent with the traits in the json. Is the website using outdated code? Am I missing something?

ogregoire:
The online API was last updated 3 years ago. A lot of cleanup, standardization and fixes have been done since then in this repo which haven't been replicated in the online API. So changing to the old trait URLs would be a regression, as the trait URLs have been updated since then.

Incognitonomous:
will the online api ever receive an update?

bagelbits:
That’s a great question. You should probably reach out to the maintainer of
that site. I’m sorry we can’t be of more help.

PR #88, opened 2019-09-26
Monster attack cleanup

bagelbits:
## Overview
This is a fix for #83, among some other things.

## Changes
- [x] Add more data to damage attacks
```json
      {
        "name": "Claw",
        "desc": "Melee Weapon Attack: +13 to hit, reach 5 ft., one target. Hit: 15 (2d6 + 8) slashing damage.",
        "attack_bonus": 13,
        "damage": [
          {
            "damage_type": "slashing",
            "damage_dice": "2d6",
            "damage_bonus": 8
          }
        ]
      },
```
- [x] Split combined data types
```json
      {
        "name": "Bite",
        "desc": "Melee Weapon Attack: +14 to hit, reach 10 ft., one target. Hit: 19 (2d10 + 8) piercing damage plus 7 (2d6) fire damage.",
        "attack_bonus": 14,
        "damage": [
          {
            "damage_type": "piercing",
            "damage_dice": "2d10",
            "damage_bonus": 8
          },
          {
            "damage_type": "fire",
            "damage_dice": "2d6",
            "damage_bonus": 0
          }
        ]
      },
```
- [x] Add in damage type urls for damage
```json
        "damage": [
          {
            "type": {
              "name": "slashing",
              "url": "http://www.dnd5eapi.co/api/damage-types/12"
            },
            "dice": "2d6",
            "bonus": 8
          }
        ]
```
- [x] Remove attack bonuses for things that have no attack
```json
      {
        "name": "Detect",
        "desc": "The dragon makes a Wisdom (Perception) check."
      }
```
- [x] Add in DC for relevant attacks
```json
      {
        "name": "Frightful Presence",
        "desc": "Each creature of the dragon's choice that is within 120 ft. of the dragon and aware of it must succeed on a DC 14 Wisdom saving throw or become frightened for 1 minute. A creature can repeat the saving throw at the end of each of its turns, ending the effect on itself on a success. If a creature's saving throw is successful or the effect ends for it, the creature is immune to the dragon's Frightful Presence for the next 24 hours.",
        "dc": {
          "type": "WIS",
          "value": 14,
          "success_type": "none"
        }
      },
```
- [x] Add in ability score urls for `dc_type`
```json
      {
        "name": "Frightful Presence",
[...]

PR #97, opened 2019-10-08
Equipment Damage Cleanup

Alex-frazer:
matching damage format of PR #88
unsure on case on:
* line 1230 - blowgun
* line 1390 - net


bagelbits (approved):
LGTM. Thank you for doing this. I'm trying to do the same with monsters.

PR #102, opened 2019-10-19
Matching Dice to string format for Damage and Equipment 

Alex-frazer:


5e-SRD-Classes.json
```diff
@@ -2,35 +2,35 @@
 	{
 		"index": 1,
 		"name": "Barbarian",
-		"hit_die": 12,
+		"hit_die": "1d12,
 		"proficiency_choices": [
 			{
 				"choose": 2,
 				"type": "proficiencies",
-				"from": [
```
bagelbits, line 10: Is there a reason we're changing `"from"` to `"skills"`?

Alex-frazer: That wasn't meant to go up yet, I need to re-do this pull request

PR #106, opened 2019-11-01
Simple JSON clean up of Monsters

bagelbits:
## Overview
Any more JSON cleanup I'm going to do is going to be easier without these polluting it.

### DIFF
--- a/5e-SRD-Monsters.json
+++ b/5e-SRD-Monsters.json
@@ -218,7 +218,9 @@
     "stealth": 7,
     "damage_vulnerabilities": [],
     "damage_resistances": [],
-    "damage_immunities": ["acid"],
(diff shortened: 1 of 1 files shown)


benjaminapetersen:
Curious if you are doing this by hand or if you are using some scripts to process this into a better format.

bagelbits:
I'm using Prettify JSON within VS Code.

benjaminapetersen:
Ah, so not a script to reformat the structure (vs by hand adjustments to all 300-ish monster entries).

bagelbits:
Yeah. Doing by hand adjustments is not a fun time if I can avoid it. But it is what I had to do with adding all of the attack info for monster abilities.

PR #107, opened 2019-11-01
Convert monster desc to lists

bagelbits:
## Overview

Instead of descriptions having newlines in them, convert them to lists of strings. This will also make the spellcasting changes easier.

### DIFF
--- a/5e-SRD-Monsters.json
+++ b/5e-SRD-Monsters.json
@@ -34,11 +34,15 @@
     "special_abilities": [
       {
         "name": "Amphibious",
-        "desc": "The aboleth can breathe air and water."
+        "desc": [
+          "The aboleth can breathe air and water."
+        ]
       },
       {
         "name": "Mucous Cloud",
-        "desc": "While underwater, the aboleth is surrounded by transformative mucus. A creature that touches the aboleth or that hits it with a melee attack while within 5 ft. of it must make a DC 14 Constitution saving throw. On a failure, the creature is diseased for 1d4 hours. The diseased creature can breathe only underwater.",
+        "desc": [
+          "While underwater, the aboleth is surrounded by transformative mucus. A creature that touches the aboleth or that hits it with a melee attack while within 5 ft. of it must make a DC 14 Constitution saving throw. On a failure, the creature is diseased for 1d4 hours. The diseased creature can breathe only underwater."
+        ],
         "dc": {
           "dc_type": {
             "name": "CON",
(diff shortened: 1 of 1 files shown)


bagelbits:
Fix for: https://github.com/adrpadua/5e-database/issues/108

Alex-frazer (approved):

benjaminapetersen:
```
"desc": [
          "The acolyte is a 1st-level spellcaster. Its spellcasting ability is Wisdom (spell save DC 12, +4 to hit with spell attacks). The acolyte has following cleric spells prepared:",
          "- Cantrips (at will): light, sacred flame, thaumaturgy",
          "- 1st level (3 slots): bless, cure wounds, sanctuary"
        ]
```
This is a bit more usable, but prob could be taken quite a bit further.

ogregoire:
I don't see how splitting in lines will improve things. By the way, why splitting lines and not sentences or words? There's no real reason for that so it's uselessly arbitrary.

The description is just that: a description, not a list of descriptions.

If you want to represent the description correctly, it should be in a text format written to show descriptions, like, for instance, HTML or Markdown. Using text without a specific format is useless.

* What does an entry starting with "-" mean?
* Why is "1st level" not in bold, just like in the SRD?

These questions are answered by using a specific format, which has a format definition. HTML, Wikitext and Markdown are the best formats for the task coming to mind. Not an arbitrary format that merely splits on lines.

benjaminapetersen:
FWIW, I would find this substantially more usable if split much futher.  Example:


```yaml
# for example sake, pasted in yaml form for readability
- special abilities: 
   - name:  Spellcasting
   - description: "The acolyte is a 1st lvl spellcaster. The spellcasting ability is....etc...."
   - ability: wisdom
   - DC: 12
   - to_hit: +4
   - spells:
       - cantrips: 
           - light
           - sacred flame
           - thaumaturgy
    - 1st level:
          - bless
          - cure wounds
          - sanctuary     
```
In other words `description` is fine for prose, but there is a lot of actual `data` in the prose that would be extremely useful from a programmatic perspective (even if just a more flexible UI).

It might take a little bit of effort, but I bet there is a fairly good data structure that would acomodate most special abilities, magical or not.

benjaminapetersen:
@ogregoire I would not embed `html` or `markdown` within, instead I would suggest structured data (like above example) that would allow for any number of flexible presentation layers.

bagelbits:
Honestly, this was just because doing the Spellcasting and Innate Spellcasting stuff is going to be less of a pain in the ass with the data in this format. Also, we have a similar format for description being a list of strings split by newlines in other parts of the database. We should consolidate on format instead of have multiple different ways to do descriptions.

ogregoire:
I've always considered the description to be a text describing the data and the data being next to the description. So if you have the spellcasting ability, you have an object representing the spellcasting **and** you have a description that is the original text version of the data being shown.

bagelbits:
That makes sense. We should then probably convert any descriptions as arrays back to strings in any of the other files.

benjaminapetersen:
Completely agree.  I have been prototyping what I posted above, keeping the `description` as `prose` (string), but writing some little scripts to extract actual data values out as structured data.

bagelbits:
We're working on the form for extracting spellcasting over here:
https://github.com/adrpadua/5e-database/issues/4

PR #109, opened 2019-11-07
Change monsters' senses to be an object

ogregoire:
Changes the senses so that it's an object instead of a simple string.

A distinction was made between "blindsight", "darkvision", "truesight" and "tremorsight" and "passive_perception": the four first map to strings to indicate a distance, while the latter maps to a score.


bagelbits:
Thank you! I've been meaning to do this!

PR #110, opened 2019-11-07
Change monsters' abilities to a list of abilities

ogregoire:
This is mostly to turn abilities into full objects. The advantage is that abilities are now linked, whereas they weren't. However, this dramatically increases the size of the monster file.

So I understand if this is not wanted and rejected.

### DIFF
--- a/5e-SRD-Monsters.json
+++ b/5e-SRD-Monsters.json
@@ -13,12 +13,38 @@
       "walk": "10 ft.",
       "swim": "40 ft."
     },
-    "strength": 21,
-    "dexterity": 9,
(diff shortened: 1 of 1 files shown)


Alex-frazer:
There is no need for this. there is another file with all this. from a readable perspective this isn't a good thing. And from a programatic perspective you can relate the ability code with the ability.

Fully linked is also fully duplicated

bagelbits:
I think I agree with @Alex-frazer on this.

ogregoire:
As I said, it's fine for me. I'll close it then.

PR #111, opened 2019-11-08
Change monsters' proficiencies to a list of proficiencies

ogregoire:
This is a much needed change. So many proficiencies and no reference to the ones in `5e-SRD-Proficiencies.json`. Well, now there is one!

### DIFF
--- a/5e-SRD-Monsters.json
+++ b/5e-SRD-Monsters.json
@@ -19,11 +19,33 @@
     "intelligence": 18,
     "wisdom": 15,
     "charisma": 18,
-    "constitution_save": 6,
-    "intelligence_save": 8,
-    "wisdom_save": 6,
-    "history": 12,
-    "perception": 10,
+    "proficiencies": [
+      {
+        "name": "Saving Throw: CON",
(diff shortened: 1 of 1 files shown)


bagelbits:
Nice! Thanks for doing this!

Though, I wonder if all of those belong in Proficiency but I think that's a different discussion.

bagelbits (approved):

ogregoire:
The SRD says the following about it, page 256:

> # Saving Throws
> 
> The Saving Throws entry is reserved for creatures that are adept at resisting certain kinds of effects. For example, a creature that isn’t easily charmed or frightened might gain a bonus on its Wisdom saving throws. Most creatures don’t have special saving throw bonuses, in which case this section is absent.
> 
> A saving throw bonus is the sum of a monster’s relevant ability modifier and its proficiency bonus, which is determined by the monster’s challenge rating (as shown in the Proficiency Bonus by Challenge Rating table).

I interpret this as the saving throws being proficiencies, which is supported by the fact that saving throws are listed under the proficiencies of each sections, and that proficiency bonuses are applied to those listed saving throws.

Issue #113, opened 2019-11-12
Remove base domain from URLs


bagelbits:
This would allow this database to more easily be used with other projects and not be tied directly to the API.

PR #124, opened 2019-12-17
Droping domain from URL

bagelbits:
## Overview

This is to address #113. I don't think there is a need for the domain to exist in the database. You can  link to things without the domain. Also, it allows for other developers  to use this database to make their own APIs.


ogregoire:
Seems good to me.

Issue #120, opened 2019-12-09
Breath Weapons could be better written


bagelbits:
This could be better written using the `choice` structure as well as showing the recharge rate. 
```json
{
        "name": "Breath Weapons (Recharge 5-6)",
        "desc": "The dragon uses one of the following breath weapons.\nFire Breath. The dragon exhales fire in an 20-foot line that is 5 feet wide. Each creature in that line must make a DC 11 Dexterity saving throw, taking 14 (4d6) fire damage on a failed save, or half as much damage on a successful one.\nSleep Breath. The dragon exhales sleep gas in a 15-foot cone. Each creature in that area must succeed on a DC 11 Constitution saving throw or fall unconscious for 1 minute. This effect ends for a creature if the creature takes damage or someone uses an action to wake it.",
        "attacks": [
          {
            "name": "Fire Breath",
            "dc": {
              "dc_type": {
                "name": "DEX",
                "url": "http://www.dnd5eapi.co/api/ability-scores/2"
              },
              "dc_value": 11,
              "success_type": "half"
            },
            "damage": [
              {
                "damage_type": {
                  "name": "Fire",
                  "url": "http://www.dnd5eapi.co/api/damage-types/4"
                },
                "damage_dice": "4d6",
                "damage_bonus": 0
              }
            ]
          },
          {
            "name": "Sleep Breath",
            "dc": {
              "dc_type": {
                "name": "CON",
                "url": "http://www.dnd5eapi.co/api/ability-scores/3"
              },
              "dc_value": 11,
              "success_type": "none"
            }
          }
        ]
      }
```

bagelbits:
Half of this is done. Just need to add the choice structure.

PR #133, opened 2020-01-02
Objectify monsters' usage limits.

ogregoire:
Make objects out of the usage limits, same as what was done for spell usages.

Partially implements #120.


bagelbits:
If you rebase, you should pick up the Travis CI changes.

5e-SRD-Monsters.json
```diff
@@ -122,8 +122,12 @@
         ]
       },
       {
-        "name": "Enslave (3/day)",
+        "name": "Enslave",
```
bagelbits, line 125: Should we drop `(3/day)` from the name? That is the full name of the ability in SRD and source material. It also keeps it  human readable. Even if it does feel silly to keep  it in there.

5e-SRD-Monsters.json
```diff
@@ -420,8 +428,13 @@
         }
       },
       {
-        "name": "Acid Breath (Recharge 5-6)",
+        "name": "Acid Breath",
         "desc": "The dragon exhales acid in a 60-foot line that is 5 feet wide. Each creature in that line must make a DC 18 Dexterity saving throw, taking 54 (12d8) acid damage on a failed save, or half as much damage on a successful one.",
+        "usage": {
+          "type": "recharge on dice",
+          "dice": "1d6",
+          "values": "5-6"
```
bagelbits: Would making this an array of integers make more sense?

5e-SRD-Monsters.json
```diff
@@ -7432,8 +7645,15 @@
         ]
       },
       {
-        "name": "Relentless (Recharges after a Short or Long Rest)",
-        "desc": "If the boar takes 7 damage or less that would reduce it to 0 hit points, it is reduced to 1 hit point instead."
+        "name": "Relentless",
+        "desc": "If the boar takes 7 damage or less that would reduce it to 0 hit points, it is reduced to 1 hit point instead.",
+        "usage": {
```
bagelbits, line 7650: For the two types of recharge, would it make more sense to structure it like so?
```json
"recharge": {
  "type": "dice",
  "dice": "1d6",
  "values: [5,6]
}
```

```json
"recharge": {
  "type": "rest",
  "rest_types": [
    "short",
    "long
  ]
}
```

5e-SRD-Monsters.json
benjaminapetersen: Yeah, reading this isn't immediately obvious to me what it means, its a bit different than other ruling.  Usually you see "3x per day" or something, I forget off the top of my head how conditional recharges work.  I have to look up to have an opinion :) 

But I do like the data extracted like this.

5e-SRD-Monsters.json
benjaminapetersen: No strong opinion, you have the object split out, which is nice.  I guess if you wanted to query across lots of things for the "enslave",its easier for exact match, in case others would have `(2/day)` or different values. Just thinking out loud, could do a substring search to avoid that, but might want it simple.

ogregoire: @bagelbits no, it's not part of the name. Just because it has the same style doesn't mean it is. The SRD list those as limited usage "notations" (see page 259). It's the same case for "costs 2 (or 3) actions" text. It's not part of the name: it's only the same style as a mnemonic for quick reference.

5e-SRD-Monsters.json
ogregoire: Yes, that makes much more sense! I should have done it that way. I'll use the expanded values (`[4, 5, 6]`) instead of a shortcut range notation that would require a full object with boundary properties.

Unless you prefer the following:

    "dice": "1d6",
    "min_value": 4

5e-SRD-Monsters.json
ogregoire: No it is not clearer. Look at the SRD (p. 259): they're defined as 3 different types of usage limitations. We should keep it that way, we should keep it as close to the source material as possible when possible.

5e-SRD-Monsters.json
bagelbits: Either  way is fine, honestly.

5e-SRD-Monsters.json
bagelbits: Fair enough!

bagelbits (approved):
LGTM

5e-SRD-Monsters.json
```diff
@@ -420,8 +428,13 @@
         }
       },
       {
-        "name": "Acid Breath (Recharge 5-6)",
+        "name": "Acid Breath",
         "desc": "The dragon exhales acid in a 60-foot line that is 5 feet wide. Each creature in that line must make a DC 18 Dexterity saving throw, taking 54 (12d8) acid damage on a failed save, or half as much damage on a successful one.",
+        "usage": {
```
benjaminapetersen, line 433: Thoughts on this might be that `usage` would describe the dragon using the ability, `recharge` may be better for the recharge.  Perhaps something like this:

```yaml
name: Acid Breath
Desc: The dragon exhales acid in a 60-foot line....
Usage:
   Dice: 12d8
   Average: 54
   Range:
      Type: line
      Distance: 60
      Unit: ft
Save:
   Ability: dexterity
   DC: 18
   Success: half damage 
Recharge:
   Type: roll
   Die: d6
   Success:
      Min: 4
```

benjaminapetersen: Sorry for the late reply.  And for using `yaml` in example :)

ogregoire: I understand, and I would agree in the absolute, but here we have to deal with the context. We are putting the SRD in a data format. And the SRD mentions "limited usage" (see page 259). We could change "usage" to "limited_usage" if you think it's clearer.

bagelbits: I have poor timing of merging things before you comment @benjaminpetersen. 😅

benjaminapetersen: clear is always good :)

And no worries, I'm not quite the most timely to respond always, tho do appreciate the effort y'all have been putting into this lately!

Issue #116, opened 2019-12-02
Monster multiattacks hard to parse


jbhaywood:
I'm wondering if it's possible to define multiattacks in a way that's more easily parsed in code. Right now, the only place that specifies all of the possible attacks is the "desc" field of the "Multiattack" action. For example, in trying to figure out which attacks are available to the "Bandit Captain" on their turn, the only information I have is, "The captain makes three melee attacks: two with its scimitar and one with its dagger. Or the captain makes two ranged attacks with its daggers." 

This is tricky to parse out programmatically. One option I can think of is to maybe add an explicit list of "action sets", where each set is considered a potential set of actions for each turn, something like:

```
"action_sets": [
 	{
		"name": "Multiattack",
		"actions": [
			{
				"name": "Scimitar",
				"count":  2,
				"type": "melee"
			},
			{
				"name": "Dagger",
				"count":  1,
				"type": "melee"
			}
		]
	},
 	{
		"name": "Multiattack",
		"actions": [
			{
				"name": "Dagger",
				"count":  2,
				"type": "ranged"
			}
		]
	}
]
```

I wouldn't mind giving this a shot if we can agree on a format.

ogregoire:
We already have a "choice" or "option" structure for similar cases, such as Half-Elf's stat option. You might want to take inspiration from those.

bagelbits:
In most (if not all) cases, `Multiattack` is just using the already defined attacks on a monster. So you probably only need the name and count.

jbhaywood:
How would you handle the -or- case for the Bandit Captain above, where the same attack can be used two different ways?

"The captain makes three melee attacks: two with its scimitar and one with its dagger. **Or** the captain makes two ranged attacks with its daggers."

bagelbits:
Maybe something like this:
```json
"attacks": [
  {
    "choose": 1,
    "type": "attacks",
    "from": [
      [
        {
          "name": "Scimitar",
          "count": 2
        },
        {
          "name": "Dagger",
          "count": 1,
          "type": "melee"
        }
      ],
      [
        {
          "name": "Dagger",
          "count": 2,
          "type": "ranged"
        }
      ]
    ]
  }
]
```

I have a feeling `type` would be more than just `melee` or `ranged` though.

jbhaywood:
Sounds good. I'm going to see how far I can get with this structure. The multiattack descriptions are pretty well formatted, but some of them have a lot going on.

bagelbits:
[...]

PR #144, opened 2020-01-15
Added objects for enumerating multiattack options, addressing issue 116.

jbhaywood:
This is the change associated with the issue I submitted: #116 

Please look it over and let me know what you think. Some of the monsters' multiattacks are almost impossible to represent in the way I've organized the data without adding more complexity. And I'm not sure that's worth it. Here are the ones that proved to be especially tricky because of special cases withing the multiattack description...

~~**Glabrezu**~~
~~_The glabrezu makes four attacks: two with its pincers and two with its fists. Alternatively, it makes two attacks with its pincers and casts one spell._~~

**Grick**
_The grick makes one attack with its tentacles. If that attack hits, the grick can make one beak attack against the same target._

**Shambling Mound**
_The shambling mound makes two slam attacks. If both attacks hit a Medium or smaller target, the target is grappled (escape DC 14), and the shambling mound uses its Engulf on it._

**Tyrannosaurus Rex**
_The tyrannosaurus makes two attacks: one with its bite and one with its tail. It can't make both attacks against the same target._

~~**Violet Fungus**~~
~~_The fungus makes 1d4 Rotting Touch attacks._~~


5e-SRD-Monsters.json
```diff
@@ -138,6 +138,18 @@
         }
       }
     ],
+    "multiattack_options": {
```
bagelbits: Should these live under the Multiattack action itself?

bagelbits:
For Glabrezu, you could probably do:
```
{
  "name": "Innate Spellcasting",
  "count": 1,
  "type": "magic"
}
```
We could probably use a different type name.

For Violet Fungus, maybe:
```
{
  "name": "Innate Spellcasting",
  "count": "1d4",
  "type": "magic"
}
```
Or potentially the dice rolling object that we use elsewhere?

And for the rest, making using of notes might be the way to go, but honestly, I'm not sure.
@ogregoire? @Alex-frazer? @benjaminapetersen?

5e-SRD-Monsters.json
jbhaywood: I could go either way on that. So whatever you think is best for the repository.

benjaminapetersen:
Worth some thought for sure.
Are you generating an empty object for creatures that do not have multiattack?  `multiattack_options: {}`. I would probably always generate the field and leave it empty rather than omit, just to ensure format consistency over time.

bagelbits:
I think that's valid, in which case, it might be easier to stick that under the Multiattack action instead.

benjaminapetersen:
Uniformity for the win!

5e-SRD-Monsters.json
bagelbits: Probably under the action.

bagelbits: Probably under the action.

jbhaywood:
I'm all for uniformity. But from what I've found while working on this, this file isn't uniform already. I had to manually merge my changes into the existing json file to keep from adding a lot of new properties. There are places where default values are written into the data, and others where they're not. Do you want to me to go ahead and make the existing structures uniform too? It would save me the manual merge step.

bagelbits:
Hmmmm that's an interesting point. At least for myself, I've been trying to keep top level keys of an object consistent across all monsters. As well as keeping common scheme shapes, like the attack damage, consistent as well. I'm open to how to improve that. Mostly trying to keep things consistent for those hitting the API, especially since changes will now immediately surface in the API. I hope that was somewhat helpful.

benjaminapetersen:
+1
Nice to catch deviance early, else will be a lot more work to fix later.

jbhaywood:
I've updated the json file based on the feedback in this thread. I've moved the multiattack options object into the multiattack action. But because multiattack is just another type of action, all action objects have the 'options' property. Maybe it'll be useful for something else down the road though.

You'll also notice tons of new properties with default values (null or 0) that weren't there before. But now all objects have the same, consistent properties. Let me know what you think.

bagelbits:
@jbhaywood Can you rebase as well? This will pick up the CI as well.

jbhaywood:
I think it's up to date now. I'm no git expert, so let me know if I did anything wrong.

bagelbits:
@jbhaywood Sorry, I keep meaning to look at this but the diff is too big for github to load. So I keep putting off pulling this down locally and inspecting the diffs.

This also makes me suspect that this should probably be split into at least two different PRs. One that handles the multiattack shapes and one that handles adding in missing fields.

caldane:
Looking at your branch I can tell the following monsters are missing multiattack options:

- Animated Armor
- Chimera
- Chuul
- Cult Fanatic
- Deva
- Doppelganger
- Dragon Turtle
- Drider
- Erinyes
- Glabrezu
- Gladiator
- Half-Red Dragon Veteran
- Horned Devil
- Hydra
- Iron Golem
- Knight
- Kraken
- Lamia
- Lizardfolk
- Merrow
- Oni
- Planetar
- Purple Worm
- Sahuagin
- Scout
- Shambling Mound
- Spy
- Tarrasque
- Thug
- Vampire Spawn
- Veteran
- Wight
- Wyvern

bagelbits:
Closing out in favor of #168, #169, and #170

PR #151, opened 2020-02-14
Add missing exhaustion description, addressing issue #150

### LINKED ISSUE #150: Exhaustion condition missing vital information
Currently the response for `/conditions/exhaustion` returns the following:

```
{
  index: "exhaustion"
  name: "Exhaustion"
  desc: ["- Exhaustion is measured in six levels."]
  url: "/api/conditions/exhaustion"
}
```

The description is incomplete.

AndruC:
I've updated exhaustion with a proposal for the response.

Given exhaustion's unique format (paragraphs of prose separated by a table), the description format is up for debate.

### DIFF
--- a/src/5e-SRD-Conditions.json
+++ b/src/5e-SRD-Conditions.json
@@ -136,7 +136,19 @@
   {
     "index": "exhaustion",
     "name": "Exhaustion",
-    "desc": ["- Exhaustion is measured in six levels."],
+    "desc": [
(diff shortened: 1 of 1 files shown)


bagelbits:
I think that's reasonable based on the current structure. We do have some inconsistency between data types on the shape of description. Sometimes it's a list and sometimes it's a string.

bagelbits (approved):
LGTM

bagelbits:
@AndruC I feel like the table structure is also something that can be iterated on as we go. Though if you wanted to get a true table structure, I'd suggest taking a Markdown approach.

Issue #81, opened 2019-08-24
Add more API languages.


ariedov:
Hey, love this repo!

I play DnD in Ukraine, and we use Russian as our primary language for the parties, so I would love to have this API available in different languages.

Maybe creating folders like `en`, `ru` and just storing the different `json` files there would be a good option?

I am ready to contribute, just don't want to deploy my own API :)

bagelbits:
That seems like a pretty interesting idea. I'm a little concerned about changing the folder structure (or in this case adding a folder structure) for the existing files. My other main concern is that we would need someone to maintain the different languages as well as make sure changes in files in one language end up propagating to the other languages. I, unfortunately, do not speak or write in Russian.

bagelbits:
It might be better to some how store the different languages together but it still wouldn't handle the divergence of trying to maintain more than one language.

bagelbits:
Were you thinking just the descriptions or all fields converted to Russian and, inevitably, other languages?

ariedov:
Yeah, pretty much. And also having the ability to request a specific language in a GET param.

benjaminapetersen:
Does the SRD exist in any other languages currently?  There is prob a risk of copyright infringement as well, validating that translations are not pulling from proprietary (D&D) info.  Just pointing that out as something to think about.

ogregoire:
I am not a lawyer.

@benjaminapetersen yes the SRD exists in other languages. According to the OGL, it is allowed to translate the SRD as long as the translation is also under OGL. So yes, you may translate the SRD if you want. You shouldn't be able to get sued for that.

But currently there are no official translation of the SRD. (By "official", I mean WotC-approved.) There are official translations of books, but not any of the SRD.

Therefore I don't see why this should be included and part of this project goal. Plus, if things are translated from the D&D books and not the SRD, we have no way to know that and this project could be hosting copyright-infringing material without anyone noticing.

It's already hard to keep non-SRD monsters, spells, races, and subclasses out of this repo, I don't see why we should take the burden of doing it in languages we don't understand.

carloslancha:
Hi guys!

I'd love to have  this API available in spanish. I understand the problems @ogregoire is pointing out and the others about changing the project structure or mantaining the language. My proposition to do it would be:

- Any pr adding a new language must be sent together with a link to the published SRD translation in order check that is following OGL. (ie: http://srd.nosolorol.com/DD5/index.html)

[...]

PR #158, opened 2020-03-25
POC: Add supoort for multi language

carloslancha:
## What does this do?
Adds support for multi language, by splitting out text into a separate file that can be hot swapped for other languages.

## How was it tested?
Unclear

This addresses #81 

Not yet.

random photo - update me


src/5e-SRD-Ability-Scores.json
```diff
@@ -1,145 +1,145 @@
...
-    ],
-    "url": "/api/ability-scores/cha"
-  }
-]
+	{
+		"index": "str",
+		"name": "LANG-KEY-str",
+		"full_name": "LANG-KEY-strength",
+		"desc": [
+			"LANG-KEY-strength-measures-bodily-power-athletic-training-and-the-extent-to-which-you-can-exert-raw-physical-force",
```
bagelbits, line 7: Should these things instead be something like `LANG-KEY-strength-description` and store that as a list instead? A key like what we have right now seems a bit silly.

bagelbits: It would probably make your hash in `en.json` be a little less ridiculious? Maybe?

ogregoire: > 
> 
> Should these things instead be something like `LANG-KEY-strength-description` and store that as a list instead? A key like what we have right now seems a bit silly.

Yes, this is totally what should happen. Having such big keys can be problematic on several fronts. The first and obvious one is that if typos are put in there, or on the contrary, fixed, we have to stick with what existed.

bagelbits: If we want to keep the values the same type for everything, then the keys can instead be `LANG-KEY-strength-description-#`, where `#` is the index.

bagelbits:
Hey, @carloslancha are you still working on this or is it on pause now?

carloslancha:
hey @bagelbits ! Sorry, I've been through a hard time (illness, work, coronastuff...). I'll take care of this when I have the time, it's being a crazy spring....

bagelbits:
@carloslancha It has been. Please! Take all the time you need. There really is no rush. I just wanted to check in with you!

bagelbits:
A few years later, we finally have support for multiple languages. Sorry for the wait!

Issue #154, opened 2020-02-28
Page numbers for monsters


howdoicomputer:
Noticed they were missing.

ogregoire:
You really want the SRD page in the monster description?

bagelbits:
@howdoicomputer Do you mean the SRD page number or the page number in the Monster Manual?

howdoicomputer:
Well, I noticed that spells have PHB page numbers as part of their structure where monsters do not. I think the Monster Manual page references would be most apt. It's helpful to have the page numbers so that you can validate that the API data is correct during a session.

I wouldn't mind submitting PRs to bring up parity. It's just something I noticed; I'm not hung up on them not being there. I'm working on a project to use the API to generate Emacs org mode documents as part of running a campaign.

bagelbits:
@howdoicomputer I think that makes sense to bring parity for that. If you want to go the extra mile, a shape like this might be better:
```
{
  title: "Monster Manual",
  page: 5
}
```

I will, however, note that the SRD monsters which are in this API can have some slight difference compared to what's in the MM. Mainly this will come across as spells. I've noticed a few spells that exist in normal 5e that are not a part of the SRD.

bagelbits:
What's the issue with just one PR that has multiple commits so you don't have to do everything in one go? That's how I've handled other things like this.

ogregoire:
I wouldn't go that route of putting page numbers at all in the SRD-based data. The page numbers aren't SRD-related data, but D&D-books specific. How can we check that the information is correct at a glance? We must own a version of the book. To me, this contadicts all that we did previously in cleaning the data.

If at all, you showed us that spells still contain data that should be cleaned.

bagelbits:
> If at all, you showed us that spells still contain data that should be cleaned.

I think that might be the correct takeaway from this. Good call, @ogregoire

bagelbits:
@ogregoire #159 That should fix it for Spells.

bagelbits:
Closing now that page numbers have been dropped from spells.

PR #159, opened 2020-04-03
Drop Book Page Numbers from Spells

bagelbits:
## Overview
This is coming off of #154. I'm going to drop the page numbers because technically all of this data is from the SRD which doesn't necessarilly match everything in the PHB.


ogregoire (approved):

bagelbits:
Oh! @ogregoire, if Discord is a thing you use, we'd love to have you on it. [Here's an invite link!](https://discord.gg/TQuYTv7) We've been recently talking about ways to allow other developers to make packages of content for things like Homebrew and easily get loaded into the DB for local development. I'd love to get your opinions on the matter.

bagelbits:
Content would probably only live in forks but still it would be nice to make it easier for others.

PR #161, opened 2020-04-09
Added random encounter json for random encounter enhancement

ghost:
Hi,

I started working on a JSON file for the random encounters enhancement and wanted to get some feedback on the structure.


ghost:
@bagelbits, mind giving some initial feedback on the structure?

ghost:
I've updated the file name and simplified the structure which would allow for simpler encounters but not really any filtering. The two examples were retrieved from this: https://old.reddit.com/r/DnDBehindTheScreen/comments/80aut5/100_random_composed_encounters_on_the_road/

ogregoire:
Where are these random encounters defined in the SRD?

ghost:
There aren't random encounters defined in the SRD as far as I know

bagelbits:
Hmmm. I think this might make more sense as something that lives solely in the API and not the Database. Though I finally looked at how to calculate Combat Encounters and wow that is really complicated.

I think some of the concern that @ogregoire has of putting this in the database and I think I share is that some of the monsters included in the encounters list might not be in the SRD. And it's unclear if that list falls under OGL at all. Also, the intent of the database is to only contain data from the SRD.

ghost:
Thanks for your response @bagelbits, I'll close out this pull request as I agree that it should most likely live in the API.

PR #165, opened 2020-04-21
Quality of Life and Multiattack Improvements for Monsters.json (Part 1 of 3)

caldane:
Monster Update from 
**Aboleth**
to
**Frost Giant**

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -11,7 +11,9 @@
     "hit_dice": "18d10",
     "speed": {
       "walk": "10 ft.",
-      "swim": "40 ft."
+      "swim": "40 ft.",
+      "burrow": null,
+      "fly": null
     },
     "strength": 21,
     "dexterity": 9,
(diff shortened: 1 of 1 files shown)


caldane:
This is a break up of the PR144 into 3 smaller PR's:
so that we can use the online git tools to manage the diff.

https://github.com/bagelbits/5e-database/pull/144

ogregoire:
I disagree with the "quality of life" point.

It might be easier for *your* language to parse everything and keep everything structured. In some other languages, specially the null hostile languages (which become more and more used, btw), having a value set to null is harder to parse.

Semantics matter here, and the absence of something is not the same as a dummy mapping.

So I would prefer to keep the data as is, put it there when it's present in the SRD, and avoid it when it's not present. Just like the SRD did: WotC didn't give legendary actions to all mobs but make it really useful only on a subset of them, they put it there when needed.

That's my opinion. Others might differ.

ogregoire:
However, I would prefer that you split your PRs in two not based on lengths but based on topic: one for the QoL and another for the multi-attack.

caldane:
@ogregoire thank you for the feedback, I would love to chat with you about the most ideal way to do this.  Are you in the api discord?

PR #167, opened 2020-04-21
Users/caldane/multiattack part1

caldane:
This is the isolated code from PR https://github.com/bagelbits/5e-database/pull/144 pertaining only to multiattack options only for **Aboleth** to **Frost Giant**


ogregoire:
Seems good. Nice job separating the two issues so quickly!

PR #168, opened 2020-04-21
Add Multiattack Improvements for Monsters.json (Part 1 of 3)

caldane:
This is the isolated code from PR #144 pertaining only to multiattack options only for **Aboleth** to **Frost Giant**

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -9418,16 +9418,19 @@
             [
               {
                 "name": "Fire Breath",
+                "note": "When Fire Breath is available",
                 "count": 1,
                 "type": "ability"
               },
               {
                 "name": "Horns",
+                "note": "When Fire Breath is available",
                 "count": 1,
                 "type": "melee"
               },
               {
                 "name": "Claws",
+                "note": "When Fire Breath is available",
                 "count": 1,
                 "type": "melee"
               }
@@ -9435,16 +9438,19 @@
             [
               {
                 "name": "Bite",
+                "note": "When Fire Breath is available",
                 "count": 1,
                 "type": "melee"
               },
               {
                 "name": "Fire Breath",
+                "note": "When Fire Breath is available",
                 "count": 1,
                 "type": "ability"
               },
               {
                 "name": "Claws",
+                "note": "When Fire Breath is available",
                 "count": 1,
                 "type": "melee"
               }
@@ -9592,6 +9598,20 @@
               {
                 "name": "Pincer",
                 "count": 2,
+                "type": "melee"
+              }
+            ],
+            [
+              {
+                "name": "Pincer",
+                "note": "When chuul is grappling",
+                "count": 2,
+                "type": "melee"
+              },
+              {
+                "name": "Tentacles",
+                "note": "When chuul is grappling",
+                "count": 1,
                 "type": "melee"
               }
             ]
@@ -13611,28 +13631,16 @@
         "name": "Multiattack",
         "desc": "The erinyes makes three attacks",
         "options": {
-          "choose": 1,
+          "choose": 3,
           "from": [
             [
               {
                 "name": "Longsword",
-                "count": 3,
+                "count": 1,
                 "type": "melee"
               }
             ],
             [
-              {
-                "name": "Longbow",
-                "count": 3,
-                "type": "ranged"
-              }
-            ],
-            [
-              {
-                "name": "Longsword",
-                "count": 2,
-                "type": "melee"
-              },
               {
                 "name": "Longbow",
                 "count": 1,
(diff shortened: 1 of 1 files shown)


bagelbits:
You'll want to rebase all three PRs against the branch you're merging against. I just updated it.

src/5e-SRD-Monsters.json
```diff
@@ -8676,7 +9394,63 @@
...
+                "count": 1,
+                "type": "melee"
+              },
+              {
+                "name": "Claws",
+                "count": 1,
+                "type": "melee"
+              }
+            ],
+            [
```
bagelbits, line 9418: For these two, I'm wondering if it makes sense to add a `note` that these are only if Fire Breath is available?

src/5e-SRD-Monsters.json
```diff
@@ -8810,7 +9584,19 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The chuul makes two pincer attacks. If the chuul is grappling a creature, the chuul can also use its tentacles once."
+        "desc": "The chuul makes two pincer attacks. If the chuul is grappling a creature, the chuul can also use its tentacles once.",
```
bagelbits: We could probably use a `note` here too? That way we can add the grapple modifier too? Maybe? I know this one is a little weird.

src/5e-SRD-Monsters.json
```diff
@@ -12602,7 +13609,91 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The erinyes makes three attacks"
+        "desc": "The erinyes makes three attacks",
+        "options": {
```
bagelbits: Nice combinatorics! :D

bagelbits: Actually, another way we could do this is:
```
{
  "choose": 3,
  "from": [
    [
      {
        "name": "Longsword",
        "count": 1,
        "type": "melee"
      }
    ],
    [
      {
        "name": "Longbow",
        "count": 1,
        "type": "ranged"
      }
    ],
    [
      {
        "name": "Variant: Rope of Entanglement",
        "count": 1,
        "type": "ability"
      }
    ]
  ]
}
```

bagelbits: That utilizes the choose format a bit better and makes it much cleaner.

caldane: OMG, that seems so obvious now, well that is a far better syntax I will improve that

jbhaywood: Is being able to choose more than one of the same option valid for multi-attacks across all monsters? The idea I was going with originally was that you always choose one of the available options until you've hit the "choose: x" limit.

bagelbits: Not for all but it is for some. There are a few where it says it has to be unique. I'm not sure the best way to handle that case though.

jbhaywood: A flag on the options object, something like "choose_unique:" true/false?

jbhaywood:
Just a general comment on the introduction of a "note" field. Doing that means this data is back to requiring code that knows how to parse english sentences, which is exactly what I was trying to get away from. This is an awesome resource, and I was using it to program monster AI for a video game, so anywhere that I couldn't rely on a known schema made that more difficult.

What about instead of an arbitrary "note" field, there was a "prerequisite" object? Something like...
```
{ 
  "type": "available_ability"
  "name: "Horror Nimbus"
}
```
or
```
{ 
  "type": "successful_attack"
  "name: "Tentacles"
}
```

And the "type" would be a small set of known values.

bagelbits:
Hmmm.... that could be interesting. @ogregoire Do you have any thoughts on this?

caldane:
@jbhaywood @bagelbits 
We could leave the note field in for a more human readable version, but I do like the idea of a `prerequisite` or `condition` property.  The problem is we are going to have to come up with things that mean something to computers because these conditions are not spelled out in the api.

Also I would suggest that to be a new issue, so we can move forward with what we have.

ogregoire:
It makes sense.

But ideally, there would be an `option` structure available like this:

    "options": {
      "choose": 1,
      "from" : [
        {
          "value": [
            {
              "name": "Bite",
              "count": 1,
              "type": "melee"
            },
            {
              "name": "Horns",
              "count": 1,
              "type": "melee"
            },
            {
              "name": "Claws",
              "count": 1,
              "type": "melee"
            }
          ]
        },
        {
          "value": [
            {
              "name": "Fire Breath",
              "count": 1,
              "type": "ability"
            },
            {
              "name": "Horns",
              "count": 1,
              "type": "melee"
            },
            {
              "name": "Claws",
              "count": 1,
              "type": "melee"
            }
          ],
          "prerequisite": {
            "type": "available_ability",
            "name": "FireBreath"
          }
        }
      ]
    }

But it starts to be very structured and the structures are numerous. I personnally like it and would tend to go that way, but I understand practicality and the repetition if that's what you prefer, and then go towards the structure.

caldane:
@jbhaywood @bagelbits @ogregoire 
So do we think that has to happen before this PR is approved.  Or can we make that a new issue and work on it when this one is done.  I am afraid the scope of this is getting hard to track.

bagelbits:
Honestly, I'm okay with keeping `note` for now as a human readable way to handle it and cut an issue to make a better programmatic way to represent the data. @jbhaywood @ogregoire ?

ogregoire:
I don't mind keeping "note" for the time being.

jbhaywood:
Sounds good to me.

bagelbits (approved):
LGTM

PR #169, opened 2020-04-22
Add Multiattack Improvements for Monsters.json (Part 2 of 3)

caldane:
This is the isolated code from PR #144 pertaining only to multiattack options only for **Gargoyle** to **Otyugh**

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -13581,7 +13581,24 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The gargoyle makes two attacks: one with its bite and one with its claws."
+        "desc": "The gargoyle makes two attacks: one with its bite and one with its claws.",
+        "options": {
+          "choose": 1,
+          "from": [
+            [
+              {
+                "name": "Bite",
+                "count": 1,
+                "type": "melee"
+              },
+              {
+                "name": "Claws",
+                "count": 1,
+                "type": "melee"
+              }
+            ]
+          ]
+        }
       },
       {
         "name": "Bite",
@@ -14060,7 +14077,19 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The ape makes two fist attacks."
+        "desc": "The ape makes two fist attacks.",
+        "options": {
+          "choose": 1,
+          "from": [
+            [
+              {
+                "name": "Fist",
+                "count": 2,
+                "type": "melee"
+              }
+            ]
+          ]
+        }
       },
       {
         "name": "Fist",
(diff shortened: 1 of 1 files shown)


src/5e-SRD-Monsters.json
```diff
@@ -16299,7 +16459,57 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The gladiator makes three melee attacks or two ranged attacks."
+        "desc": "The gladiator makes three melee attacks or two ranged attacks.",
+        "options": {
```
bagelbits, line 16463: Would nested choices make sense here? I'm honestly not sure but I'm open to ideas for it. Also, this could just be the easiest approach.

src/5e-SRD-Monsters.json
```diff
@@ -17235,7 +17445,26 @@
...
+            [
+              {
+                "name": "Tentacles",
+                "count": 1,
+                "type": "melee"
+              }
+            ],
+            [
+              {
+                "name": "Beak",
```
bagelbits, line 17461: We should probably add a `note` field here as well since this only works if the first one hits.

src/5e-SRD-Monsters.json
```diff
@@ -17960,7 +18218,19 @@
...
+        "desc": "The veteran makes two longsword attacks. If it has a shortsword drawn, it can also make a shortsword attack.",
+        "options": {
+          "choose": 1,
+          "from": [
+            [
+              {
+                "name": "Longsword",
+                "count": 2,
+                "type": "melee"
+              }
```
bagelbits, line 18230: We can probably add the shortword attack with a `note` field.

src/5e-SRD-Monsters.json
```diff
@@ -19638,7 +20063,38 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The golem makes two melee attacks."
+        "desc": "The golem makes two melee attacks.",
+        "options": {
+          "choose": 1,
```
bagelbits, line 20068: This is probably a good candidate as well for the `"choose": 2,` format.

src/5e-SRD-Monsters.json
```diff
@@ -20935,7 +21475,84 @@
     "actions": [
       {
         "name": "Multiattack",
-        "desc": "The lizardfolk makes two melee attacks, each one with a different weapon."
+        "desc": "The lizardfolk makes two melee attacks, each one with a different weapon.",
+        "options": {
+          "choose": 1,
```
bagelbits, line 21480: Geeze, I wish we had a good schema shape for "Choose two distinct". This is probably fine though?

src/5e-SRD-Monsters.json
```diff
@@ -22842,7 +23599,29 @@
...
       {
         "name": "Multiattack",
-        "desc": "The nalfeshnee uses Horror Nimbus if it can.  It then makes three attacks: one with its bite and two with its claws."
+        "desc": "The nalfeshnee uses Horror Nimbus if it can.  It then makes three attacks: one with its bite and two with its claws.",
+        "options": {
+          "choose": 1,
+          "from": [
+            [
+              {
+                "name": "Horror Nimbus",
```
bagelbits, line 23608: I'd add a `note` field here for if it's available?

bagelbits (approved):
LGTM

PR #181, opened 2020-05-07
Update 5e-SRD-Equipment-Categories.json

cachance7:
Missing hyphens on several urls

## What does this do?
Fixes url references to equipment

## How was it tested?
Via api web app to ensure fixed urls exist and old urls do not

No

random photo - update me


bagelbits (approved):
Nice catch!

PR #185, opened 2020-05-09
Removing redundant traits from Gnome Subrace

### LINKED ISSUE #184: Gnome race traits duplicated on Rock Gnome subrace
Rock Gnome has Darkvision and Gnome Cunning traits listed:

https://github.com/bagelbits/5e-database/blob/d3ff744c6e03a31074790e6ee195981e3893a34b/src/5e-SRD-Subraces.json#L244-L251

Per SRD these are Gnome race traits, they are not specific to the Rock Gnome subrace.

Gnome race data does have them too.

bagelbits:
## What does this do?
Removes Darkvision and Gnome Cunning from Rock Gnome as it's already in the Gnome race.

## How was it tested?
CI

This resolves #184 

image

### DIFF
--- a/src/5e-SRD-Subraces.json
+++ b/src/5e-SRD-Subraces.json
@@ -242,14 +242,6 @@
     "language_options": {},
     "racial_traits": [
       {
-        "name": "Darkvision",
-        "url": "/api/traits/darkvision"
-      },
-      {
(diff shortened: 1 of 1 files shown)


cachance7:
LGTM!

PR #188, opened 2020-05-09
Fixed Thief subclass flavor

Rarst:
## What does this do?

Thief subclass has Sorcerer's flavor string, probably copy/paste error or something.


bagelbits (approved):
LGTM!

PR #191, opened 2020-05-26
Update 5e-SRD-Subclasses.json with the correct levels links

EkkaDokhari:
## What does this do?
Changed the incorrect links to the class levels to correct ones.

## How was it tested?
Auto checked the replaced links

No


bagelbits (approved):
LGTM!

PR #210, opened 2020-07-03
Use dice notation for damage

ogregoire:
## What does this do?
1. Merge the `"damage_dice"` and `"damage_bonus"` fields into one `"damage_dice"` field with [standard dice notation](https://en.wikipedia.org/wiki/Dice_notation).
2. Remove the damage for the net, since its damage is stroked through in the SRD.

No, not yet

random photo - update me

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -6494,7 +6494,7 @@
               "name": "Bludgeoning",
               "url": "/api/damage-types/bludgeoning"
             },
-            "damage_dice": "0d4+1"
+            "damage_dice": "1"
           }
         ]
       }
@@ -7158,7 +7158,7 @@
     "actions": [
       {
         "name": "Bite",
-        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one creature. Hit: 1 piercing damage.",
+        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one creature. Hit: 1 piercing damage.",
         "attack_bonus": 0,
         "damage": [
           {
@@ -7166,7 +7166,7 @@
               "name": "Piercing",
               "url": "/api/damage-types/piercing"
             },
-            "damage_dice": "0d4+1"
+            "damage_dice": "1"
           }
         ]
       }
@@ -8884,7 +8884,7 @@
     "actions": [
       {
         "name": "Claws",
-        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one target. Hit: 1 slashing damage.",
+        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one target. Hit: 1 slashing damage.",
         "attack_bonus": 0,
         "damage": [
           {
@@ -8892,7 +8892,7 @@
               "name": "Slashing",
               "url": "/api/damage-types/slashing"
             },
-            "damage_dice": "0d4+1"
+            "damage_dice": "1"
           }
         ]
       }
@@ -10437,7 +10437,7 @@
     "actions": [
       {
         "name": "Claw",
-        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one target. Hit: 1 bludgeoning damage.",
+        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one target. Hit: 1 bludgeoning damage.",
         "attack_bonus": 0,
         "damage": [
           {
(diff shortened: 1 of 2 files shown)


ogregoire:
Please note that I'm fully aware of the rule of the critical which doubles the damage dice without bonus (or triples if a Balor inflicts the damage with its longsword), however the data is as it is in the book: "NdX+Y". I believe we should keep it as is and leave the separation of the dice and bonus to the implementation, just like a lot of the rest. Writing a parser is a rather easy task and we could provide examples of parsing rules in the wiki (I gladly could).

bagelbits:
I think adding examples of parsing rules in the wiki is a really great idea. I know that I've also gotten repeated questions on calculating hit points for monsters. Let me post this up on the discord as well so people know that this is coming.

bagelbits:
My main concern is that this might be considered a breaking change? Which doesn't mean we shouldn't do it. But I'm not sure if now is a good time to introduce the idea of versioning.

src/5e-SRD-Monsters.json
```diff
@@ -9073,16 +8884,15 @@
     "actions": [
       {
         "name": "Claws",
-        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one target. Hit: 1 slashing damage.",
+        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one target. Hit: 1 slashing damage.",
```
AndruC: Add back in

src/5e-SRD-Monsters.json
```diff
@@ -22532,16 +22114,15 @@
     "actions": [
       {
         "name": "Bite",
-        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one target. Hit: 1 piercing damage.",
+        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one target. Hit: 1 piercing damage.",
```
AndruC: Add +0 attack mod back in

src/5e-SRD-Monsters.json
```diff
@@ -10650,16 +10437,15 @@
     "actions": [
       {
         "name": "Claw",
-        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one target. Hit: 1 bludgeoning damage.",
+        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one target. Hit: 1 bludgeoning damage.",
```
AndruC: Fix

src/5e-SRD-Monsters.json
```diff
@@ -27693,16 +27188,15 @@
     "actions": [
       {
         "name": "Bite",
-        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one target. Hit: 1 piercing damage.",
+        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one target. Hit: 1 piercing damage.",
```
AndruC: Fix

src/5e-SRD-Monsters.json
```diff
@@ -7310,16 +7158,15 @@
     "actions": [
       {
         "name": "Bite",
-        "desc": "Melee Weapon Attack: +0 to hit, reach 5 ft., one creature. Hit: 1 piercing damage.",
+        "desc": "Melee Weapon Attack:  to hit, reach 5 ft., one creature. Hit: 1 piercing damage.",
```
AndruC: Fix

AndruC (changes requested):
Regression: +0 to hit removed from monsters

ogregoire:
@AndruC Nice catch! I fixed all the issues you mentioned and a few others I saw while fixing those.

AndruC:
I'm going to leave the approval to @bagelbits at this point, given the comments on breaking changes and versioning

bagelbits:
@ogregoire I'll probably merge this in later this evening and close out the versioning stuff I did. My one ask is that either in this PR or a followup, that you add a section to the wiki for parsing rules.

ogregoire:
@bagelbits I can do that before next week, and this PR can stay open until that is done as well, if you like.

ogregoire:
Just a small notification to say that [the grammar was added to the wiki](https://github.com/bagelbits/5e-database/wiki/Dice-Notation). Is this what you expected, @bagelbits ?

bagelbits:
I guess that'll work. Never used ANTLR before though. :D

bagelbits (approved):
LGTM

PR #211, opened 2020-07-06
Scorching Ray writeup, leading spaces in desc fix

antihumorous:
## What does this do?
This update changes some leading spaces in spell descriptions for consistency. Antimagic Field's description was updated for consistency with how other spells handled bolded description headers. It also updates the description for Scorching Ray, which was transcribed incorrectly.

## How was it tested?
I ran a regex search and replace and did a quick look over. I also tested it by rendering the descriptions with my chat bot.

Not that I know of.

random photo - update me


src/5e-SRD-Spells.json
```diff
@@ -9889,11 +9896,11 @@
     "name": "Scrying",
     "desc": [
       "You can see and hear a particular creature you choose that is on the same plane of existence as you. The target must make a wisdom saving throw, which is modified by how well you know the target and the sort of physical connection you have to it. If a target knows you're casting this spell, it can fail the saving throw voluntarily if it wants to be observed.",
-      "Knowledge &amp; Save Modifier\n",
+      "Knowledge &amp; Save Modifier",
```
bagelbits: Not on you but should this be `&amp;`?

src/5e-SRD-Spells.json
```diff
@@ -9889,11 +9896,11 @@
...
       "Secondhand (you have heard of the target) +5",
       "Firsthand (you have met the target) +0",
       "Familiar (you know the target well) -5",
-      "Connection &amp; Save Modifier\n",
+      "Connection &amp; Save Modifier",
```
bagelbits: Same here?

bagelbits:
@antihumorous Small comment but otherwise looks great!

src/5e-SRD-Spells.json
antihumorous: Probably a Javascript parsing thing. I believe it's all the same, but we can change them to strict ampersands if you prefer. Either way, I'm good if you wanna merge this. :)

bagelbits: Yeah, do you mind converting them to strict ampersands? Then it's more consistent. I'll merge it in after you do that.

antihumorous:
All done!

bagelbits (approved):
LGTM!

PR #212, opened 2020-07-06
Update 5e-SRD-Spells.json with damage, heal, saving throws and area of effect

LetsRoleRPG:
## What does this do?
This PR adds some data to spells, so it's easier to use in a VTT. This data as been crowdsourced to some people in the Let's Role community : ChrisFly, Cortez, Eril2212, Indemmar, Jaggerwock, Ocleos, Shiva, Tenshi and TruePika.

Beware that this PR is far from perfect, but I think is a good start.

## How was it tested?
Exporting the data using PHP was kind of annoying since the json_encode function does not output in the format you use. But I checked the whole file with a diff tool, and other than removing an unused line, the existing data should not have changed.

https://github.com/bagelbits/5e-database/issues/209

random photo - update me


ogregoire:
You've done a pretty good job, @LetsRoleRPG ! Congrats! :-)

However I have to admit that there are things that I don't like in this PR. But rest assured, it's not you: it's the data and the wording of the spells that actually isn't consistent from one spell to another, and by using the `"desc"` field in the `"dc"` field, these are more visible than in the full text.

I guess, this is as good as we can get for a first commit and I'm glad we're here to see how big of a job this actually is.

The things I don't like are roughly the following:

* the description sometimes references the caster, sometimes the target.
* the structure doesn't allow yet for hooks for what happens later (example: acid arrow)
* The absence of a `"dc_fail'` is annoying, even though we understand that it's always the damage or the conditions applied that are seen. (example: bestow curse)
* When multiple types of damage are applied, it's not possible to see which is which (example: flame strike).

So I guess, this is a very good first draft and we could use it to expand and go further the way we dataify the spells.

ogregoire:
However I believe we can already integrate this as I didn't see any issue anywhere that would make this PR not mergeable. Most of the spells are already correct and those that aren't, it's because they are special and require a better attention.

LetsRoleRPG:
Please do not merge now, I just noticed I forgot the "attack_type" field.

LetsRoleRPG:
All good :)

src/5e-SRD-Spells.json
```diff
@@ -16,6 +16,13 @@
     "concentration": false,
     "casting_time": "1 action",
     "level": 2,
+    "attack_type": "ranged",
+    "damage": {
```
bagelbits, line 20: Should this also include the damage as well?

bagelbits:
I agree with @ogregoire on pretty much all the same points. I don't love it but it's going in the right direction and works as a good first draft. I don't really see any reason other than my small comment to not merge it in and continue to iterate on the design.

bagelbits:
@LetsRoleRPG I have one comment on your change. If you address it, I'm happy to merge it in.

LetsRoleRPG:
Here you go :)

bagelbits (approved):
LGTM!

PR #215, opened 2020-07-14
Versioning DB

bagelbits:
## What does this do?
**Note: Do not merge this in until the API PR has also been approved. This PR should be merged in first out of the two though.**

This is a two part PR (the other part is in the API repo), that introduces the concept of versioning to the API and DB. This is a result of #210, which would introduce a breaking a change to the API since it alters how we show damage dice in the shape. The intent.

## How was it tested?
Local and CI

None.

image


ogregoire:
I'm not a fan of the versioning of the database for several reasons that I will outline here.

Before dwelling on the reasons, I can understand that the data would require a version mechanism. It might be important to be able to tell when the data incompatibly changed and when it did not.

The idea here is that we're in a continuous delivery system (as far as I understand the 5e-srd-api project, every change done here is picked up and published), but no one ever said they want us to be accountable for the data. Yet, every commit is a release in itself, with everything that comes with it: new version, having to decide whether it's minor or major. That's such a big hassle for a hobby project, and so much responsibility. At no point are we actually releasing a file that makes different versions coexist at the same time (unless I'm mistaken, of course, but that's not what I see).

It's unclear who the version is for. What message do we want to convey and for who regarding the version. We barely have any feedback on the project except for corrections so we don't hear "hey, that version was better". The communication coming with versions is absent. 

Another issue I have is that there are absolutely no plans to make several versions coexist (and there shouldn't be any). The latest version is and should always be the one shown. It would make no sense to have the version of the data without the dice notation (v1) to be available with the version of the data with the dice notation (v2). What this means is that when v2 is released, v1 goes to the bin. The two will not live together at any point. So basically, using a version name like "latest" will always be right and will always be the best version. So since all versions are the latest, what's even the point of versioning.

Also, we don't want to maintain two separate versions: we want to improve the structure and the data and that's it. Well, at least I personnally don't want to maintain separate versions: I only want to improve the data. Once the dice notation is integrated, I consider it done and if somebody says "I'm using that old commit as baseline and want to have data kept up to date on that baseline", I'm going to say no. When we make changes, we make changes in the main branch, we don't keep several branches in parallel.

To me, the idea of versioning of this project is a requirement for goodfeel, but it's absolutely not required. I'm questioning the need for versioning in this project because it gives us absolutely all the inconveniences coming with versioning but none of the advantages.

While I've laid out the reasons why I'm against versioning here, I'm not closed to it: I merely ask to open the conversation and become convinced.

bagelbits:
Honestly, I appreciate your input as always @ogregoire. I also appreciate the time you put into writing all of your thoughts down. I hadn't really considered just not doing it and I thought it was just something that I *must* do. I think all of your points are 100% valid and honestly, I don't want to support multiple versions and you're right with how we wouldn't really have more than one active at a time (I think my original plan was to have an old version active for a month before turning it down).

Thank you for giving me a really good alternative perspective to this idea. I'm going to leave this open for a bit in case anyone else has any strong feels, but as of right now, I'm 100% leaning towards dropping the idea of versioning entirely.

AndruC:
The most compelling reason for versioning requires maintaining access to prior versions: preventing consumers of the API from breaking silently when changes are made to latest. 

It would be a lot of work, and require those versions to coexist. Not worth it IMO when consumers can always use their own local copy.

bagelbits:
> It would be a lot of work, and require those versions to coexist. Not worth it IMO when consumers can always use their own local copy.

@AndruC I feel similarly.

bagelbits:
Cool. I'm going to close this out and never think about versioning again. Thanks all!

PR #216, opened 2020-07-19
Fix spear Damage

ragnork2:
## What does this do?
The spear damage was listed as 1d4, spears do 1d6 damage

## How was it tested?
Eyeballs

No

random photo - update me


bagelbits (approved):
LGTM!

Issue #223, opened 2020-08-01
Proficiencies should container a pointer to the referenced resource


dgoetsch:
Given a proficiency, I should be able to look up the referenced resource.

There should be an additional field on the proficiency resource:

```
"references": [{
  "name": "Dagger",
  "type": "equipment",
  "url": "/api/equipment/dagger"
}]
```

Examples:
* [the daggers proficiency](https://www.dnd5eapi.co/api/proficiencies/daggers) should contain a reference to [the dagger equipment](https://www.dnd5eapi.co/api/equipment/dagger)
* [the martial weapons proficiency](https://www.dnd5eapi.co/api/proficiencies/martial-weapons) should contain a reference to the [martial weapons equipment category](https://www.dnd5eapi.co/api/equipment-categories/martial-weapons)
* [the acrobatics proficiency](https://www.dnd5eapi.co/api/proficiencies/skill-acrobatics) should contain a reference to the [acrobatics skill](https://www.dnd5eapi.co/api/skills/acrobatics)

dgoetsch:
I'll take this one, but can't assign to myself

bagelbits:
Assigned to you! Also, closing out since you did an awesome job of handling it!

PR #224, opened 2020-08-01
#223 add reference array to proficiencies resources

dgoetsch:
## What does this do?
Adds array of references to the subject of the proficiency resource.

## How was it tested?
Verified all new reference urls against the api via script.
Verified the type matches the url resourceType for all new reference objects.

snippet:

```kotlin
        val urlBase = "http://localhost:8099"
        val mapper = jacksonObjectMapper()
        runBlocking {
            val proficiencies = mapper
                    .readValue<List<Proficiency>>(File("proficiencies.json").readText())
            proficiencies.map { proficiency -> proficiency.references?.map {
                assert(it.url.startsWith("/api/${it.type}/"))
                val response = khttp.get("$urlBase${it.url}")
                assertEquals(response.statusCode, 200, "Path ${it.url} was unsuccessful")
                assertNotNull(response.jsonObject)
            } }
        }
```
Isssue #223

### DIFF
--- a/src/5e-SRD-Proficiencies.json
+++ b/src/5e-SRD-Proficiencies.json
@@ -25,8 +25,13 @@
 		"name": "Warlock"
 	}],
 	"races": [],
-	"url": "/api/proficiencies/light-armor"
-}, {
+	"url": "/api/proficiencies/light-armor",
+	"references": [{
+		"url": "/api/equipment-categories/light-armor",
+		"type": "equipment-categories",
+		"name": "Light armor"
+	}]
+  }, {
 	"index": "medium-armor",
 	"type": "Armor",
 	"name": "Medium armor",
@@ -44,15 +49,25 @@
 		"name": "Ranger"
 	}],
 	"races": [],
-	"url": "/api/proficiencies/medium-armor"
-}, {
+	"url": "/api/proficiencies/medium-armor",
+	"references": [{
+		"url": "/api/equipment-categories/medium-armor",
+		"type": "equipment-categories",
+		"name": "Medium armor"
+	}]
+  }, {
 	"index": "heavy-armor",
 	"type": "Armor",
 	"name": "Heavy armor",
 	"classes": [],
 	"races": [],
-	"url": "/api/proficiencies/heavy-armor"
-}, {
+	"url": "/api/proficiencies/heavy-armor",
+	"references": [{
+		"url": "/api/equipment-categories/heavy-armor",
+		"type": "equipment-categories",
+		"name": "Heavy armor"
+	}]
+  }, {
 	"index": "all-armor",
 	"type": "Armor",
 	"name": "All armor",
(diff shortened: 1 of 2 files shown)


src/5e-SRD-Proficiencies.json
```diff
@@ -862,22 +1319,37 @@
...
 	"index": "land-vehicles",
 	"type": "Vehicles",
 	"name": "Land Vehicles",
 	"classes": [],
 	"races": [],
-	"url": "/api/proficiencies/land-vehicles"
-}, {
+	"url": "/api/proficiencies/land-vehicles",
+	"references": [{
+		"url": "/api/equipment-categories/tack-harness-and-drawn-vehicles",
```
dgoetsch: I'm not too sure about this one, since it include saddles, barding, and bit and briddle. However, this appears to be the best option.

src/5e-SRD-Proficiencies.json
```diff
@@ -895,8 +1367,13 @@
 		"name": "Ranger"
 	}],
 	"races": [],
-	"url": "/api/proficiencies/saving-throw-str"
-}, {
+	"url": "/api/proficiencies/saving-throw-str",
+	"references": [{
+		"url": "/api/ability-scores/str",
+		"type": "ability-score",
```
dgoetsch: I looked for a saving throw resource and didn't see one. So, I reasoned that the saving throw resource is the ability score resource.

Is this correct? Should this just be an empty array?

src/5e-SRD-Proficiencies.json
```diff
@@ -64,99 +79,169 @@
...
+	}]
+  }, {
 	"index": "shield",
 	"type": "Armor",
 	"name": "Shield",
 	"classes": [],
 	"races": [],
-	"url": "/api/proficiencies/shield"
-}, {
+	"url": "/api/proficiencies/shield",
```
bagelbits: It seems weird that there is both `Shield` and `Shields`? Should these be actually combined instead?

src/5e-SRD-Proficiencies.json
```diff
@@ -590,246 +865,413 @@
...
+  }, {
 	"index": "dragonchess-set",
 	"type": "Gaming Sets",
 	"name": "Dragonchess set",
 	"classes": [],
 	"races": [],
-	"url": "/api/proficiencies/dragonchess-set"
-}, {
+	"url": "/api/proficiencies/dragonchess-set",
+	"references": []
```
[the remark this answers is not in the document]
bagelbits: As far as I can tell, Dragonchess isn't in the SRD either so it should just be removed.

src/5e-SRD-Proficiencies.json
```diff
@@ -590,246 +865,413 @@
...
+  }, {
 	"index": "three-dragon-ante-set",
 	"type": "Gaming Sets",
 	"name": "Three-Dragon Ante set",
 	"classes": [],
 	"races": [],
-	"url": "/api/proficiencies/three-dragon-ante-set"
-}, {
+	"url": "/api/proficiencies/three-dragon-ante-set",
+	"references": []
```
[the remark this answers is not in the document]
bagelbits: This also isn't in the SRD so any reference to it should also be removed.

src/5e-SRD-Proficiencies.json
bagelbits: I think that makes the most sense, unless you want to add a land vehicles Equipment Category, which I think would be completely reasonable in this case. Especially since we have a `Waterborne vehicles` category.

src/5e-SRD-Proficiencies.json
bagelbits: ¯\_(ツ)_/¯
Honestly, I think that makes sense.

bagelbits (changes requested):
I think this makes sense just a few small things to do!

src/5e-SRD-Proficiencies.json
dgoetsch: This is a symptom of the data.  In dnd5 there is only one shield, but the APi exposes both a shield equipment resource and a shields equipment-categories resource. Since proficiencies can be on specific items, or on equipment categories both proficiencies are required here.

Moreover, all of the classes point at "shields" even though I think they should point at shield.

I can replace all references to /api/equipment-categories/shield to /api/equipment/shields and remove the shields equipment-category

src/5e-SRD-Proficiencies.json
dgoetsch: I think I will make a category

src/5e-SRD-Proficiencies.json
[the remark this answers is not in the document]
dgoetsch: acknowledged

src/5e-SRD-Proficiencies.json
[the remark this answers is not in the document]
dgoetsch: acknowledged

src/5e-SRD-Proficiencies.json
dgoetsch: Ahhhhh! But all the weapon proficiencies are plural. I think I'll leave the proficiency plural, point at the singular equipment item, and get rid of the shields equipment category

bagelbits (approved):
LGTM!

PR #226, opened 2020-08-04
Dbrefresh with node

clemarescx:
## What does this do?
Runs the database initialisation / restore script with NodeJS instead of *nix shell to make the script cross-platform (thus now including Windows).

The new script looks in the `src` directory for all JSON files with names following the pattern *5e-SRD-**\<datatype\>**.json*, where **\<datatype\>** is the collection name in Camel-Kebab case (e.g.: Ability-Scores) and the mongodb collection name is generated by setting **\<datatype\>** to lowercase. 

## How was it tested?
This was tested on Windows 10 with NodeJS v12.18.3 and MongoDB v4.2
MONGODB_URI was always set to the example URI: `mongodb://localhost/5e-database`

Not that I could see at first glance over issue titles 🙂


package-lock.json
```diff
@@ -0,0 +1,6467 @@
+{
```
bagelbits: Please use `yarn` instead of `npm`. There is already a yarn lockfile.

scripts/db-refresh.js
```diff
@@ -0,0 +1,72 @@
...
+} catch (e) {
+    console.error('could not execute mongoimport - make sure the directory containing mongoimport is visible in your $PATH environment variable')
+    process.exit(1);
+}
+
+const json_db_dir = 'src';
+const json_db_collection_prefix = '5e-SRD-';
+const json_data_pattern = `\\b${json_db_collection_prefix}(.+)\.json\\b`;
+
+const regex = new RegExp(json_data_pattern);
```
bagelbits, line 32: This will end up picking up the Test file. Though in all honesty, I'm not even sure why that file exists.

bagelbits: I might just remove the file, honestly.

bagelbits: I made a separate PR to fix this: https://github.com/bagelbits/5e-database/pull/227

bagelbits (changes requested):
Really just the switch from `npm` to `yarn` needs to change. Otherwise, this looks fine!

package-lock.json
clemarescx: Sorry I'm not sure what you meant by that, did you mean I should delete the *package-lock.json* file?

scripts/db-refresh.js
clemarescx: I edited the PR description text to remove the mention of the *test* JSON file + collection.

package-lock.json
bagelbits: Yes please! Sorry, I thought you also ended up installing new packages as well. But it looks like you weren't.

bagelbits: Or apparently, I can?

bagelbits (approved):
LGTM!

PR #228, opened 2020-08-05
Spacing changes, added templates.json

antihumorous:
This fixes some spacing and string array structure issues with spells.json. Further, it adds a new file templates.json to a templates folder in the root directory.

Testing was done with a json validator.

random photo - update me


bagelbits (approved):
I think that looks pretty good! I'll leave this open for a day to see if anyone else has any other comments before I merge it in.

ogregoire:
Thanks for spaces fix and the templates.json.

However I would strongly have them separated in two different issues: I have remarks on templates.json but none on the spaces fix.

My remarks about the templates.json are the following:

* Why does it exist? What is the goal you expect from having that file? To me, it makes no sense having it now that most of the job is done, now that we're more likely to fine-tune the data rather than describing it. This looks like a template to add new stuff. We've kind of already proven that the data is complete. Not entirely structured, but everything is in there. So I don't get why this is useful. Maybe you have some explanation that I have not thought of, but at this point, I believe new data shouldn't be added to the SRD as data. As usual, I try to keep my point of view as open as possible, so please convince me it's useful.
* Assuming you convinced me above, why is it only one file? I would rather have one file per type. That's easier than having to look in one unique file where everything is stored.
* Why is the template inconsistent? I see that sometimes, the placeholder data is `${class.name}`, sometimes it's `0` (such as in spell slots). I don't get why the placeholders simply exist. If you want to validate, why not use appropriate tools such as JSON Schema Validator (the first result when I googled for "how to validate a JSON structure"). To me this templates.json file looks like the wrong tool for the job. Plus, a tool such as a schema validator can sometimes be reversed to add new elements to the structure. Not always but often.

I would strongly suggest these questions are answered before we merge this PR. Also, I would much prefer the space fix to be in a separate PR.

kylrs:
@ogregoire many of these questions are answered in our discussion about this PR on the project's Discord server.

`templates.json` is designed to facilitate the addition of new content, including non-core materials, homebrew, and new SRD content, should there ever be any.

There could be a case made for providing seperate template files for each resource, I agree.

In some cases, such as `url` fields, values need to be interpolated into a specific format. This is why the implementation uses the JavaScript format string notation, which is both human and machine friendly. _Numerical_ placeholder values are 0. This is so that it is clear that the required value is numeric, and harder than most other numbers to be confused with an actual value that should remain untouched.

These are all things discussed in the Discord server prior the the PR being opened. If you have any better ideas, I'm sure they'd be welcome.

ogregoire:
If things are discussed elsewhere, it might be a good idea to summarize them in the PR when dropped here, and to provide a link to the full discussion.

Also it doesn't change the fact that PR should be atomic and that fixing text and spells shouldn't be done at the same place as providing an entire new feature. So splitting should still occur. So that one discussion doesn't block the other texts.

bagelbits:
@ogregoire Does have a very good point in keeping these separate.

PR #230, opened 2020-08-06
Refactor starting equipment choices

fergcb:
## What does this do?
This PR refactors starting equipment resources, replacing an arbitrary number of `choice_n` attributes with a single `choices` array for each class' starting equipment entry.

This achieves several things:
-  **Computer-friendliness** - The existing format requires the data consumer to read the number `choices_to_make`, and then iterate over the attributes of the pattern `choice_n` - this is effectively a long-winded implementation of a JSON array; hence, this PR replaces these attributes with a single array, allowing for much more straight-forward reading by the client code.
- **Code Reusability** - By using `choice` structures for each set of choices to be made, instead of arrays, the logic involved in selecting an item can be reused for selecting *which* choice should be made by the user.
- **Remove unnecessary choices** - Previously, in an attempt to remain consistent, where there was a choice between 1 specific item, or any item from an equipment category, *both* options would be supplied in a `choice` structure. This PR removes unnecessary `choice` structures, by replacing them with single `equipment` objects, or arrays of `equipment` objects, where more than one item *must* be chosen, e.g. when a ranged weapon comes with ammunition.

## How was it tested?
ESLint to validate JSON. 

too many choices

### DIFF
--- a/src/5e-SRD-StartingEquipment.json
+++ b/src/5e-SRD-StartingEquipment.json
@@ -4,71 +4,45 @@
 		"url": "/api/classes/barbarian",
 		"name": "Barbarian"
 	},
-	"starting_equipment": [{
-		"item": { "url": "/api/equipment/explorers-pack", "name": "Explorer's Pack"}, "quantity": 1}, {
-		"item": { "url": "/api/equipment/javelin", "name": "Javelin"}, "quantity": 4}
+	"starting_equipment": [
+		{ "equipment": { "url": "/api/equipment/explorers-pack", "name": "Explorer's Pack" }, "quantity": 1 },
+		{ "equipment": { "url": "/api/equipment/javelin", "name": "Javelin" }, "quantity": 4 }
 	],
 	"starting_equipment_options": [
 		{
 			"choose": 1,
-			"type": "choice",
+			"type": "equipment",
 			"from": [
+				{ "equipment": { "name": "Greataxe", "url": "/api/equipment/greataxe" }, "quantity": 1 },
 				{
-					"choose": 1,
-					"type": "equipment",
-					"from": [{
-						"item": { "url": "/api/equipment/greataxe", "name": "Greataxe"}, "quantity": 1}]
-				}, {
-					"choose": 1,
-					"type": "equipment",
-					"from": [{
-						"item": { "url": "/api/equipment/battleaxe", "name": "Battleaxe"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/flail", "name": "Flail"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/glaive", "name": "Glaive"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/greataxe", "name": "Greataxe"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/greatsword", "name": "Greatsword"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/halberd", "name": "Halberd"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/lance", "name": "Lance"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/longsword", "name": "Longsword"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/maul", "name": "Maul"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/morningstar", "name": "Morningstar"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/pike", "name": "Pike"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/rapier", "name": "Rapier"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/scimitar", "name": "Scimitar"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/shortsword", "name": "Shortsword"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/trident", "name": "Trident"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/war-pick", "name": "War pick"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/warhammer", "name": "Warhammer"}, "quantity": 1}, {
-						"item": { "url": "/api/equipment/whip", "name": "Whip"}, "quantity": 1}]
+					"equipment_option": {
+						"choose": 1,
+						"type": "equipment",
+						"from": {
+							"equipment_category": {
+								"name": "Martial Melee Weapons",
+								"url": "/api/equipment-categories/martial-melee-weapons"
+							}
+						}
+					}
 				}
 			]
 		}, {
 			"choose": 1,
-			"type": "choice",
+			"type": "equipment",
 			"from": [
+				{ "equipment": { "name": "Handaxe", "url": "/api/equipment/handaxe" }, "quantity": 2 },
 				{
-					"choose": 1,
-					"type": "equipment",
-					"from": [{
-						"item": { "url": "/api/equipment/handaxe", "name": "Handaxe"}, "quantity": 2}]
-				}, {
-					"choose": 1,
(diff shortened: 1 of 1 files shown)


src/5e-SRD-StartingEquipment.json
```diff
@@ -6,60 +6,73 @@
...
-			"item": { "url": "/api/equipment/sickle", "name": "Sickle"}, "quantity": 1}, {
-			"item": { "url": "/api/equipment/spear", "name": "Spear"}, "quantity": 1}, {
-			"item": { "url": "/api/equipment/crossbow-light", "name": "Crossbow, light"}, "quantity": 1}, {
-			"item": { "url": "/api/equipment/dart", "name": "Dart"}, "quantity": 1}, {
-			"item": { "url": "/api/equipment/shortbow", "name": "Shortbow"}, "quantity": 1}, {
-			"item": { "url": "/api/equipment/sling", "name": "Sling"}, "quantity": 1}]
-	}],
+		"item": { "url": "/api/equipment/javelin", "name": "Javelin"}, "quantity": 4}
+	],
+	"choices": [
```
ogregoire: Should be named "starting_equipment_options"

fergcb: Done!

src/5e-SRD-StartingEquipment.json
```diff
@@ -6,60 +6,73 @@
...
+			"from": [
+				{
+					"choose": 1,
+					"type": "equipment",
+					"from": [{
+						"item": { "url": "/api/equipment/greataxe", "name": "Greataxe"}, "quantity": 1}]
+				}, {
+					"choose": 1,
+					"type": "equipment",
+					"from": [{
```
ogregoire: The data is "one greataxe or any martial melee weapon". So we should have a choice of "one greataxe" and "one martial melee weapon":

```
{
  choose: 1
  type: equipment
  from: [
    {
      equipment: {
        url: "...",
        name: Greataxe
      }
      quantity: 1
    },
    {
      equipment_option: {
        choose: 1
        type: equipment
        from: {
          equipment_category: martial-melee-weapon
        }
      }

    }
  ]
}
```

(this was roughly written, but the idea is there.)

fergcb: Are we happy here to lose the `quantity` attribute for the nested options?

ogregoire: The option says "choose 1 from", so that acts as quantity, right?

I'm open to the discussion.

src/5e-SRD-StartingEquipment.json
```diff
@@ -4,844 +4,678 @@
...
+								"name": "Martial Weapons",
+								"url": "/api/equipment-categories/martial-weapons"
+							}
+						}
+					}
+				}
+			]
+		}, {
+			"choose": 1,
+			"type": "choice",
```
bagelbits: Should this be `"type": "equipment",`?

fergcb: Yep. Good catch! I'm on it.

src/5e-SRD-StartingEquipment.json
```diff
@@ -4,844 +4,678 @@
...
+					}
+				}
+			]
+		}, {
+			"choose": 1,
+			"type": "equipment",
+			"from": [
+				{ "equipment": { "url": "/api/equipment/component-pouch", "name": "Component pouch" }, "quantity": 2 },
+				{
+					"equipment_option": {
```
bagelbits, line 543: Do we need this or can it be combined into the parent choice? I think this is also in a few other places too.

src/5e-SRD-StartingEquipment.json
```diff
@@ -4,844 +4,678 @@
...
+							}
+						}
+					}
+				}
+			]
+		}, {
+			"choose": 1,
+			"type": "equipment",
+			"from": [
+				{ "equipment": { "url": "/api/equipment/component-pouch", "name": "Component pouch" }, "quantity": 2 },
```
bagelbits: Is the quantity supposed to be 2? I think this might be repeated in a few other places as well.

fergcb: This must be an artefact from the old data. Will update it.

src/5e-SRD-StartingEquipment.json
bagelbits: Huh.... in the SRD, it's just `Arcane Focus` and not all of these choices. Weird. That looks like it was weirdly pulled from the SRD:
```
An arcane focus is a special item—an orb, a crystal, a rod, a specially constructed staff, a wand-like length of wood, or some similar item—designed to channel the power of arcane spells. A sorcerer, warlock, or wizard can use such an item as a spellcasting focus.
```

bagelbits: Huh: https://github.com/bagelbits/5e-database/blob/master/src/5e-SRD-Equipment.json#L1924-L2012

bagelbits: Don't worry about fixing the Arcane Focus issue. That should be a separate PR.

bagelbits: I'll throw one together once this gets merged in.

bagelbits:
Other than those changes, I think this looks great. I like the use of Equipment categories as well.

src/5e-SRD-StartingEquipment.json
fergcb: When applying my initial changes, I did wonder whether to combine these. I chose to leave them as-is, so as to better reflect the source text of the SRD, as well has the Player's Handbook. It would certainly make sense to merge them in order to reduce the number of choices a user has to make, but I wonder if that should be prioritised over aligning with the source texts, for familiarity's sake. Thoughts?

fergcb: Ah alright. Will leave it for now.

bagelbits: Hrmmmm. Okay yeah. I see the issue. Especially since each kind of Arcane Focus has a different weight and cost. Well this is weird. This might make more sense as an Equipment Category? Maybe?

bagelbits: @ogregoire Thoughts?

ogregoire: Arcane Focus :

> Arcane Focus. An arcane focus is a special item — an orb, a crystal, a rod, a specially constructed staff, a wand -­‐‑ like length of wood, or some similar item — designed to channel the power of arcane spells. A sorcerer, warlock, or wizard can use such an item as a spellcasting focus.

There's no other description or category, so I'd just use "Arcane Focus".

bagelbits: @ogregoire 
The catch is that arcane focus isn't really listed in the cost weight table, instead it's orb, crystal, rod, staff, or wand. If you look on page 69 of the SRD.

bagelbits: And the Equipment table has those listed as well instead of Arcane Focus.

fergcb: Worth opening a new issue/PR before this discussion gets any longer?

bagelbits: My best thought is to setup an equipment category for Arcane Focus and include those items in the category instead, which more accurately encapsulates how the data is actually protrayed?

bagelbits: @fergcb Yes please.

bagelbits: Please link the issue here! :D

fergcb: Here's the issue: https://github.com/bagelbits/5e-database/issues/234

bagelbits (approved):
I think since we broke out the arcane/druidic focus into a separate issue. This looks pretty good! @ogregoire How do you feel about merging this in as is?

ogregoire:
@bagelbits I'm fine with it as I merged it 😁

bagelbits:
Nice

PR #235, opened 2020-08-09
New equipment categories for spellcasting foci and holy symbols

### LINKED ISSUE #234: Starting equipment arcane foci are repeated/redundant
In the Starting Equipment endpoint, every choice thats allow an Arcane focus to be chosen list the equipment items recommended by the SRD.

This raises two key issues:
1. the list is repeated several times. This means any changes to the list have to be made in several places, which is a hit to maintainability.
2. the list contains only the equipment items explicitly mentioned by the SRD as possible foci:

​
```
An arcane focus is a special item - 
an orb, a crystal, a rod, a specially constructed staff, a wand-like length of wood, or some similar item - 
designed to channel the power of arcane spells. A sorcerer, warlock, or wizard can use such an item as a spellcasting focus.
```

The first issue could be resolved by moving the suggested arcane foci into an `arcane-foci` equipment category. This would allow easier maintenance and addition of homebrew data, and would make the choice more consistent with other starting equipment options.

The second issue is trickier; given that *any* item can be a focus, with suitable fluff to back it up, an equipment category cannot contain all possible arcane foci. The question then becomes, how likely is it that such a non-standard focus will be chosen by the user, and is that event worth supporting within the data itself, if possible?

It would be most straightforward, and most inline with the SRD source text, to offer a selection of arcane foci from an equipment category containing the items referenced in the SRD description of am arcane focus. If a user desires an alternative focus, they may swap one of the default options for another item manually, after choosing their starting equipment.

*This issue was raised after discussion on PR #230 in this [comment thread](https://github.com/bagelbits/5e-database/pull/230#discussion_r467213978)*

fergcb:
## What does this do?
- Creates three new equipment categories, "Arcane Foci", "Druidic Foci" and "Holy Symbols", containing the suggested equipment items that may be used for each purpose, according to the SRD.
- Removes arcane foci, druidic foci and holy symbols from the "Standard Gear" category (though they remain in the "Adventuring Gear" category).
- Replaces the lists of foci equipment items included in starting equipment choices with references to the appropriate equipment category, to match similar references to weapon categories.

## How was it tested?
JSON validated with ESLint.

Resolves #234

arcane focus

### DIFF
--- a/src/5e-SRD-Equipment-Categories.json
+++ b/src/5e-SRD-Equipment-Categories.json
@@ -254,26 +254,6 @@
 				"name": "Antitoxin (vial)"
 			},
 			{
-				"url": "/api/equipment/crystal",
-				"name": "Crystal"
-			},
-			{
-				"url": "/api/equipment/orb",
-				"name": "Orb"
-			},
-			{
-				"url": "/api/equipment/rod",
-				"name": "Rod"
-			},
-			{
-				"url": "/api/equipment/staff",
-				"name": "Staff"
-			},
-			{
-				"url": "/api/equipment/wand",
-				"name": "Wand"
-			},
-			{
 				"url": "/api/equipment/backpack",
 				"name": "Backpack"
 			},
@@ -368,22 +348,6 @@
 			{
 				"url": "/api/equipment/crowbar",
 				"name": "Crowbar"
-			},
-			{
-				"url": "/api/equipment/sprig-of-mistletoe",
-				"name": "Sprig of mistletoe"
-			},
-			{
-				"url": "/api/equipment/totem",
-				"name": "Totem"
-			},
-			{
-				"url": "/api/equipment/wooden-staff",
-				"name": "Wooden staff"
-			},
-			{
-				"url": "/api/equipment/yew-wand",
-				"name": "Yew wand"
 			},
 			{
 				"url": "/api/equipment/emblem",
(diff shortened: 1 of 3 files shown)


src/5e-SRD-Equipment-Categories.json
```diff
@@ -656,6 +620,42 @@
 			{
 				"url": "/api/equipment/scholars-pack",
 				"name": "Scholar's Pack"
+			},
+			{
```
bagelbits, line 624: Should the holy symbols be added here?

src/5e-SRD-Equipment-Categories.json
```diff
@@ -2280,5 +2232,72 @@
 			}
 		],
 		"url": "/api/equipment-categories/waterborne-vehicles"
+	},
+	{
+		"index": "arcane-foci",
```
bagelbits, line 2237: For these three, should we update the corresponding equipment to point to this instead of Adventuring Gear? Or update the `gear category`?

bagelbits:
Two small comments but this looks fantastic!

src/5e-SRD-Equipment-Categories.json
fergcb: Good spot. The `gear_category` should updated to be plural, to match the equipment categories' names.

The `equipment_category` of the various foci should remain "Adventuring Gear", I think, to match the way it's done in other types of adventuring gear that have their own equipment sub-categories (e.g. kits, artisan's tools, vehicles).

Perhaps a good PR for the future would swap the `gear_category` string for a reference to the equipment sub-category? For now, I'll switch gear categories to plural names.

src/5e-SRD-Equipment-Categories.json
fergcb: The holy symbols are all already in the `adventuring-gear` equipment category, if that's what you're suggesting?

bagelbits: Oh excellent. I think I was confused why we moved the other things around but it doesn't really matter.

bagelbits: Oh excellent. I think I was confused why we moved the other things around but it doesn't really matter.

src/5e-SRD-Equipment-Categories.json
bagelbits: I think that would be a good future PR!

src/5e-SRD-Equipment-Categories.json
fergcb: Yeah, my fault really. I accidentally removed the foci from `adventuring-gear` instead of `standard-gear`. By the time I realised, it was too late to <kbd>CTRL</kbd>+<kbd>Z</kbd>, so just whacked them back in.

src/5e-SRD-Equipment-Categories.json
fergcb: Sorted.

bagelbits (approved):
LGTM

PR #239, opened 2020-08-11
Add all subclass Levels and features

### LINKED ISSUE #176: Some subclasses missing from Levels
Currently, we only include the subclasses for Bard and Barbarian. We're missing the SRD subclasses for everything else.

I'm not sure if we're missing these subclasses in other places as well. It's probably worth checking out.

fergcb:
## What does this do?
- Removes `features` array from Subclass objects, as features are given by subclass levels
- Adds all subclass levels from the SRD as Level objects, containing feature lists

## How was it tested?
JSON validated with ESLint, tested with Jest.

Resolves #176 

defect or feature

### DIFF
--- a/src/5e-SRD-Levels.json
+++ b/src/5e-SRD-Levels.json
@@ -8407,13 +8407,13 @@
   },
   {
     "level": 10,
-    "feature_choices": [],
-    "features": [
+    "feature_choices": [      
       {
-        "name": "Additional Fighting Style",
-        "url": "/api/features/additional-fighting-style"
-      } 
+        "name": "Choose: Additional Fighting Style",
+        "url": "/api/features/choose-additional-fighting-style"
+      }
     ],
+    "features": [],
     "spellcasting": {},
     "subclass_specific": {},
     "class": {
@@ -8674,13 +8674,13 @@
   },
   {
     "level": 3,
-    "feature_choices": [],
-    "features": [
+    "feature_choices": [
       {
         "name": "Choose: Hunter's Prey",
         "url": "/api/features/choose-hunters-prey"
-      } 
+      }
     ],
+    "features": [],
     "spellcasting": {},
     "subclass_specific": {},
     "class": {
@@ -8696,13 +8696,13 @@
   },
   {
     "level": 7,
-    "feature_choices": [],
-    "features": [
+    "feature_choices": [
       {
         "name": "Choose: Defensive Tactics",
         "url": "/api/features/choose-defensive-tactics"
-      } 
+      }
     ],
+    "features": [],
     "spellcasting": {},
     "subclass_specific": {},
     "class": {
@@ -8718,13 +8718,13 @@
   },
   {
     "level": 11,
-    "feature_choices": [],
-    "features": [
+    "feature_choices": [
       {
         "name": "Choose: Multiattack",
         "url": "/api/features/choose-multiattack"
       } 
     ],
+    "features": [],
     "spellcasting": {},
     "subclass_specific": {},
     "class": {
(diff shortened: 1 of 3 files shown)


src/5e-SRD-Levels.json
```diff
@@ -8158,5 +8154,974 @@
...
+    "class": {
+      "name": "Cleric",
+      "url": "/api/classes/cleric"
+    },
+    "subclass": {
+      "name": "Life",
+      "url": "/api/subclasses/life"
+    },
+    "url": "/api/subclasses/life/levels/1",
+    "index": 248
```
bagelbits, line 8182: Not for this PR, but we might as well swap out the index for `${class.name || subclass.name}-${level}`

src/5e-SRD-Levels.json
```diff
@@ -8158,5 +8154,974 @@
...
     "url": "/api/subclasses/lore/levels/14",
     "index": 247
+  },
+  {
+    "level": 1,
+    "feature_choices": [],
+    "features": [
+      {
+        "name": "Bonus Proficiency",
+        "url": "/api/features/bonus-proficiency"
```
bagelbits, line 8164: I hope this doesn't bite us in the ass later. But it's fine for now.

src/5e-SRD-Levels.json
```diff
@@ -8158,5 +8154,974 @@
...
+    "url": "/api/subclasses/land/levels/2",
+    "index": 253
+  },
+  {
+    "level": 6,
+    "feature_choices": [],
+    "features": [
+      {
+        "name": "Land's Stride",
+        "url": "/api/features/lands-stride"
```
bagelbits: This URL doesn't exist. I believe you want `/api/features/druid-lands-stride`

bagelbits: You may want to do a once over and make sure all of your feature urls exist. I wish the CI covered this check but it doesn't yet.

src/5e-SRD-Levels.json
fergcb: Agreed. I kept thinking that while I was working on this.

src/5e-SRD-Levels.json
fergcb: What do you mean, exactly?

src/5e-SRD-Levels.json
fergcb: I thought I had checked all of them! I'll fix this one and double check.

src/5e-SRD-Levels.json
bagelbits: Mind cutting another issue! :D

src/5e-SRD-Levels.json
bagelbits: If we ever have another class with Bonus Proficiency but it's not say "Heavy Armor". Luckily Bard has a similar feature but it's called "Bonus Proficiencies".

fergcb: Ah I see. I didn't add any features in this PR, only used existing ones. Perhaps one for a separate issue?

src/5e-SRD-Levels.json
fergcb: Done! I've written a script to resolve all the reference URLs and they're all coming back positive :)

bagelbits (approved):
LGTM!

ogregoire:
I wouldn't have merged it yet because the features are all mixed. I'll try to sort them as part of #241

fergcb:
@ogregoire What do you mean, "all the features are mixed"?

ogregoire:
I was under the impression that the order of class features and subclass features was not consistent accross the files. The order is actually consistent, so I don't know why I was so sure of it that I posted this comment. My bad.

bagelbits:
It's okay!

fergcb:
Features weren't touched by this PR, with the exception of renaming Holy Nimbus, anyway. If they aren't/weren't in the right order, that ought to be addressed by a different PR (your general "tidying up" one, perhaps).

PR #242, opened 2020-08-13
Switch Starting Equipment to string index

### LINKED ISSUE #237: Switch starting equipment index to class string instead of number
This is in line with what we do everywhere else.

bagelbits:
## What does this do?
This switches the number index to a string index which is the just the class name as there is a 1:1 relationship between Class and Starting Equipment

This requires https://github.com/bagelbits/5e-srd-api/pull/78 to be merged immediately after this one for a clean cutover.

## How was it tested?
CI

This resolves #237 

image

### DIFF
--- a/src/5e-SRD-Classes.json
+++ b/src/5e-SRD-Classes.json
@@ -68,7 +68,7 @@
       }
     ],
     "starting_equipment": {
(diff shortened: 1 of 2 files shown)


fergcb (commented):
Looks good to me!

PR #249, opened 2020-08-27
Standardise APIResource structures

### LINKED ISSUE #243: Standardise APIResource structures
Throughout the database, objects contain references to other objects in the database. Most commonly, references have the following structure:

```json
{
    "name": "Name of Referenced Resource",
    "url": "/api/.../..."
}
```

However, elsewhere, we several variations on this structure, for example, in some attributes of Class objects, we have cases like this:

```json
"starting_equipment": {
    "url": "/api/starting-equipment/1",
    "class": "Barbarian"
},
```

In this case, the `class` attribute is pointless, as `starting_equipment` is a property of the Class object, which already contains it's own name. Only the `url` attribute is useful here.

When querying collections using the API, we get an array of references of the following format:
```
{
    "index": "resource-index",
    "name": "Resource Name",
    "url": "/api/resource/url"
}
```

I believe this is the ideal structure for references, and should be implemented across the database, for these reasons:
- The `url` is provided in order to fetch the resource from the API - this is the most important part of the reference - it needs to stay
- `index` is useful for managing the data client-side. For example, it can be used to fetch the resource from a local cache, as a key for the resource in a collection, etc. 
    - This is useful as the index does not always correspond with the final element of the `url`, so cannot be reliably parsed from the `url` attribute. e.g. `/api/classes/monk/levels` - the index of the referenced resource is nowhere to be seen in the URL.
- The `name` is provided as a human-readable identifier of the resource. This comes in handy if displaying a list of objects to the user in a meaningful way, without having to fetch each object from the API.
    - `name` should be optional, as there are scenarios where it is less useful or non-applicable, for example Starting Equipment and Level strctures lack a `name` attribute.

fergcb:
## What does this do?
This PR makes sure that all NamedAPIResource structures contain *at least* `name`, `index` and `url` attributes. In cases where attributes besides `url` were redundant or misleading, such as in `class_levels` of Class resources, this PR replaces the entire object with a single string containing the URL of the referenced object.

Where necessary, additional attributes have been allowed, such as `bonus` in ability score bonuses, or `type` in proficiencies. In future, these could be unpacked from the structure to separate the reference from other information relating to the parent object, though the complication of the added layer of nesting may be unwelcome.

## How was it tested?
JSON validated with ESLint, data tested with Jest, loaded into mongo & API tests run to confirm.

Resolves #243 

I don't always use standard memes


src/5e-SRD-Classes.json
```diff
@@ -9,76 +9,84 @@
         "type": "proficiencies",
         "from": [
           {
-            "url": "/api/proficiencies/skill-animal-handling",
-            "name": "Skill: Animal Handling"
+            "index": "skill-animal-handling",
+            "name": "Skill: Animal Handling",
```
bagelbits, line 13: Not related to this PR, but I wonder if these should just have `Skill:` drop from their name and index.

src/5e-SRD-Classes.json
```diff
@@ -9,76 +9,84 @@
...
-    "starting_equipment": {
-      "url": "/api/starting-equipment/barbarian",
-      "class": "Barbarian"
-    },
-    "class_levels": {
-      "url": "/api/classes/barbarian/levels",
-      "class": "Barbarian"
-    },
+    "starting_equipment": "/api/starting-equipment/barbarian",
+    "class_levels": "/api/classes/barbarian/levels",
```
bagelbits, line 84: I left a comment on the open issue about these.

src/5e-SRD-Classes.json
fergcb: Probably a good shout

src/5e-SRD-Races.json
```diff
@@ -5,9 +5,10 @@
     "speed": 25,
     "ability_bonuses": [
       {
+        "index": "con",
         "name": "CON",
-        "url": "/api/ability-scores/con",
-        "bonus": 2
+        "bonus": 2,
```
bagelbits: It's weird that the bonus is in here with this. But probably not something that needs to be done in this PR. `bonus` at this level makes sense but it feels like the rest of the data should be keyed off something?

src/5e-SRD-Subclasses.json
```diff
@@ -1,1225 +1,1786 @@
...
-	},
-	"name": "Evocation",
-	"subclass_flavor": "Arcane Tradition",
-	"desc": ["You focus your study on magic that creates powerful elemental effects such as bitter cold, searing flame, rolling thunder, crackling lightning, and burning acid. Some evokers find employment in military forces, serving as artillery to blast enemy armies from afar. Others use their spectacular power to protect the weak, while some seek their own gain as bandits, adventurers, or aspiring tyrants."],
-	"subclass_levels": {
-		"url": "/api/subclasses/evocation/levels"
-	},
-	"url": "/api/subclasses/evocation"
-}]
+[
```
bagelbits: Can you make a separate PR with the reformatting so it's not muddled with these changes?

src/5e-SRD-Spells.json
```diff
@@ -4447,12 +5232,16 @@
...
+      "The target\u00e2\u20ac\u2122s size doubles in all dimensions, and its weight is multiplied by eight. This growth increases its size by one category\u00e2\u20ac\u201dfrom Medium to Large, for example. If there isn\u00e2\u20ac\u2122t enough room for the target to double its size, the creature or object attains the maximum possible size in the space available. Until the spell ends, the target also has advantage on Strength checks and Strength saving throws. The target\u00e2\u20ac\u2122s weapons also grow to match its new size. While these weapons are enlarged, the target\u00e2\u20ac\u2122s attacks with them deal 1d4 extra damage.",
```
bagelbits: This looks like some artifacts that aren't supposed to be here. They might show up in other places too.

bagelbits: If you search the whole codebase for unicode characters with `\u`, you should be able to catch any places this happened.

src/5e-SRD-Monsters.json
```diff
@@ -21,29 +21,34 @@
     "charisma": 18,
     "proficiencies": [
       {
+        "index": "saving-throw-con",
         "name": "Saving Throw: CON",
-        "url": "/api/proficiencies/saving-throw-con",
-        "value": 6
```
bagelbits, line 26: Also, not in this PR, but I'm not sure how I feel about having the value in here.

src/5e-SRD-Monsters.json
```diff
@@ -235,32 +253,32 @@
             {
               "name": "Light",
               "level": 0,
-              "url": "http://www.dnd5eapi.co/api/spells/180"
```
bagelbits, line 238: Huh. None of those are valid URLs. I must have missed those when I transferred spells over to string indices.

src/5e-SRD-Monsters.json
```diff
@@ -10503,12 +11231,17 @@
         "spellcasting": {
           "level": 4,
           "ability": {
+            "index": "5",
             "name": "WIS",
-            "url": "http://www.dnd5eapi.co/api/ability-scores/5"
+            "url": "/api/ability-scores/5"
```
bagelbits, line 11236: This url also doesn't match. Might want to do a regex for:
```
"url": "/api/[a-z\-]+/[0-9]+"
```
That way we can see if anything else got missed by accident.

bagelbits: This can be done in a separate PR though.

src/5e-SRD-Monsters.json
bagelbits: This can be done in a separate PR though.

bagelbits:
Also, is this the image you're using:
image

src/5e-SRD-Races.json
fergcb: I agree. That's something I already mentioned in the original PR comment.

src/5e-SRD-Subclasses.json
fergcb: It's literally just whitespace changes, does it really warrant its own PR?

src/5e-SRD-Spells.json
fergcb: Oops, must be a side effect of something in my toolchain.

src/5e-SRD-Monsters.json
fergcb: Again, see the original PR comment.

fergcb:
> Also, is this the image you're using:

@bagelbits yes it is. I didn't bother trying to fix it, as it didn't seem important at all to the PR.

src/5e-SRD-Races.json
bagelbits: Apparently my reading comprehension is terrible.

src/5e-SRD-Subclasses.json
bagelbits: Let me take another look. It could be fine. Though usually, I just do a separate PR when I do automatic formatting changes since they're basically no-op.

src/5e-SRD-Subclasses.json
```diff
@@ -1,1225 +1,1786 @@
...
+      "url": "/api/classes/cleric"
+    },
+    "name": "Life",
+    "subclass_flavor": "Divine Domain",
+    "desc": [
+      "The Life domain focuses on the vibrant positive energy--one of the fundamental forces of the universe--that sustains all life. The gods of life promote vitality and health through healing the sick and wounded, caring for those in need, and driving away the forces of death and undeath. Almost any non-evil deity can claim influence over this domain, particularly agricultural deities, sun gods, gods of healing or endurance, and gods of home and community."
+    ],
+    "spells": [
+      {
+        "prerequisites": [
```
bagelbits: I was definitely confused by this structure, but I'm not sure of a better way and also outside of scope for this PR.

src/5e-SRD-Subclasses.json
bagelbits: @fergcb I can mostly make it out but it isn't pleasant to review. It's not just whitespace changes. There's a bunch of new lines as well, which the normal `hide white spaces` setting doesn't  get rid of.

bagelbits: I'll just merge in the format myself so you don't have to bother with another PR.

bagelbits: Done! Just merge in the changes from master.

bagelbits: ....unless that just added a bunch of conflicts for you.

fergcb: I was including line breaks as whitespace. My point being that no structures have changed, only text formatting. There's not really anything to review in that specific diff (unless you were looking at the entire PR diff, rather than individual commit?). It would be just as challenging/simple to review as a separate PR.

Nevermind. If you've already made the change to master I'll figure out how to replay my commits over it in a few days when I'm next available.

bagelbits: That looks fine! I'll undo what I did so no conflicts.

src/5e-SRD-Spells.json
fergcb: Fixed!

src/5e-SRD-Subclasses.json
fergcb: Yeah, this is definitely one for future discussion

bagelbits (approved):
LGTM!

PR #254, opened 2020-09-04
Add martial weapons choice to shield option

### LINKED ISSUE #253: Paladin starting equipment has an incomplete choice option
So, one of starting equipment choices allows the player to choose either `a shield and any martial weapon` or `two martial weapons`, but in API first choice doesnt include weapon, only shield is specified.

fergcb:
## What does this do?
Fixes the first equipment option in Paladin's starting equipment to include a choice of one martial weapon with the shield.

## How was it tested?
Linted, Jested, loaded into local mongo instance. 

Resolves #253 - Thanks @Freedzio!

N/A

weapons

### DIFF
--- a/src/5e-SRD-StartingEquipment.json
+++ b/src/5e-SRD-StartingEquipment.json
@@ -691,14 +691,29 @@
         "choose": 1,
         "type": "equipment",
         "from": [
-          {
-            "equipment": {
-              "index": "shield",
-              "name": "Shield",
-              "url": "/api/equipment/shield"
+          [
+            {
+              "equipment": {
+                "index": "shield",
+                "name": "Shield",
+                "url": "/api/equipment/shield"
+              },
+              "quantity": 1
             },
-            "quantity": 1
-          },
+            {
+              "equipment_option": {
+                "choose": 1,
+                "type": "equipment",
+                "from": {
+                  "equipment_category": {
+                    "name": "Martial Weapons",
+                    "index": "martial-weapons",
+                    "url": "/api/equipment-categories/martial-weapons"
+                  }
+                }
+              }
+            }
+          ],
           {
             "equipment_option": {
               "choose": 2,


src/5e-SRD-StartingEquipment.json
```diff
@@ -691,14 +691,29 @@
...
+                "from": {
+                  "equipment_category": {
+                    "name": "Martial Weapons",
+                    "index": "martial-weapons",
+                    "url": "/api/equipment-categories/martial-weapons"
+                  }
+                }
+              }
+            }
+          ],
```
ogregoire, line 716: Why do we have an array and an object as elements of the `from` clause? I prefer consistency. Is there any way to reconciliate a unique structure instead of a difference in structure?

fergcb: I can't think of a decent way that maintains the structure of the choices presented in the SRD.

You could of course accept that regardless of what the player chooses, they will always have to take one martial weapon, and this would allow you to de-nest the structure a little; by adding an additional choice for a single martial weapon, we could "simplify" this (pseudo-JSON):
```
{
  "choose": 1,
  "from": [
    [
      { item: shield x1 },
      { choice: any martial weapon x 1 }
    ],
    { choice: any martial weapon x2 }
  ]
}
```
to this:
```
{
  "choose": 1,
  "from": [
    { item: shield x1 },
    { choice: any martial weapon x1 }
  ]
},
{
   "choose": 1,
   "from": { any martial weapon }
}
```

The latter is less deeply nested, but still has the issue of mixing choices and item references, and is further from what the SRD actually says, so could become confusing. I personally think it's a bad idea to try to make all elements of the "from" list the same type, as it would mean *radically* changing the structure of the data, further obfuscating the source content. Also, it's trivial to check the shape of each object is, and handle it appropriately.

If you have any ideas about how to achieve your preference without obfuscating the actual decisions that are offered by the SRD, I would be interested to hear them, perhaps in a new issue. As it stands, this is change is consistent with the current data.

Freedzio: If I may add something, IMO @bagelbits is right. The choosables can take really different shapes and the equipment choosing process is quite different for every class. Sometimes its an array of items, sometimes you can choose a specific item or ANY item from an array, sometimes u can choose an array of items or any three items from other array.

fergcb: @Freedzio, you mean @fergcb ;)

Freedzio: oh, my bad :D I was alt-tabbing through lots of windows and his name flashed for a number of times, kinda scrawled into my brain :P

ogregoire: I was actually thinking about something like this:

```
{
  choose : 1,
  from: [
    [
      { shield },
      { choose 1 from martial weapons }
    ],
    [
      { choose 2 from martial weapons }
    ]
  ]
}
```

I'm at work so I can't dwell much, but does this make sense to you? Do you think this is an agreeable alternative?

fergcb: I mean, it could be a good compromise. Again, I would suggest opening a new issue to discuss it. I'm gonna resolve this review because it's not directly relevant to this PR.

fergcb: I'm not introducing this with this PR... It happens throughout the starting equipment documents that the elements of a Choice's "from" array don't have the same shape.

fergcb: See for example in the current Cleric starting equipment, we have this choice structure:
```
{
    "choose": 1,
    "type": "equipment",
    "from": [
      [
        {
          "equipment": {
            "index": "crossbow-light",
            "name": "Crossbow, light",
            "url": "/api/equipment/crossbow-light"
          },
          "quantity": 1
        },
        {
          "equipment": {
            "index": "crossbow-bolt",
            "name": "Crossbow bolt",
            "url": "/api/equipment/crossbow-bolt"
          },
          "quantity": 20
        }
      ],
      {
        "equipment_option": {
          "choose": 1,
          "type": "equipment",
          "from": {
            "equipment_category": {
              "index": "simple-weapons",
              "name": "Simple Weapons",
              "url": "/api/equipment-categories/simple-weapons"
            }
          }
        }
      }
    ]
  }
```
You can see that the first option in the choice is an array of two equipment items (much like the Paladin's shield and martial weapon), and that the second is a choice (much like the Paladin's any two martial weapons).

I welcome a case for modifying the structure of the StartingEquipment documents (though I find little reason to do so, as the current approach is both open to many trivial computational methods for resolution, and is faithful to the SRD), but to block this PR for "inconsistency" with the *existing* data and structures is nonsensical, as far as comparable structures in the data demonstrate.

bagelbits: That's fair enough.

bagelbits: If we want to rework this, it should be applied holistically in a separate PR. This matches what we're doing as is.

PR #257, opened 2020-09-13
Add spells URL to Class documents

### LINKED ISSUE #250: Expose the `/api/classes/${class}/spells` endpoint in the classes resource.
The list of class spells is available using the API, but isn't discoverable. Adding a `class_spells` attribute with the value `/api/classes/${class}/spells` would expose the spell list, so that the URL doesn't have to be hard-coded or computed by the client.

fergcb:
## What does this do?
This PR exposes the `/api/classes/${class_index}/spells` endpoint, making class spell lists discoverable by the client.

## How was it tested?
Locally (eslinted, jested, loaded into local mongodb instance)

Resolves #250 

Yes! [Here's the PR in the API repo](https://github.com/bagelbits/5e-srd-api/pull/93)

image

### DIFF
--- a/src/5e-SRD-Classes.json
+++ b/src/5e-SRD-Classes.json
@@ -90,6 +90,7 @@
       }
     ],
     "spellcasting": {},
+    "spells": "/api/classes/barbarian/spells",
     "url": "/api/classes/barbarian"
   },
   {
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
LGTM!

Issue #187, opened 2020-05-09
Should proficiency–granting trait be explicit in race data?


Rarst:
Mechanically races do not grant proficiencies, but they can grant _traits_ that grant proficiencies.

E.g. Elf _race_ has Keen Senses _trait_ that grants Perception _skill proficiency_.

However Elf race data has no mention of Keen Senses, it just lists Perception proficiency. In reverse Keen Sense trait does list Elf race.

While this is perfectly fine on practical level, it kind of omits a step in mechanics.

What should be the best case for race data in such cases?
1. List only granted proficiencies (current)
2. List both granting traits and granted proficiencies
3. List only granting traits

Dropping existing proficiencies field is likely not preferable, but I think adding traits explicitly might be, for comprehensiveness?

bagelbits:
I could see #2 making sense. @ogregoire, any thoughts?

ogregoire:
In my view, a race is a set of traits. It should therefore contains those traits or references to them. The other way around makes much less sense to me.

As @Rarst says, the races list their skill proficiencies directly without mentioning their traits.

I believe that the API currently makes the shortcut of adding whatever it can to structured content and leaves the rest to random, unstructured traits.

It would definitely be better to indeed have both the full list of traits and their references to the traits file, but also structure as much as possible to integrate the result of those traits in the race.

Therefore I believe that number 2 is also the best way to do so.

PR #269, opened 2020-09-19
Added the missing dwarf trait.

mikolajg-tt:
The trait Tool Proficiency wasn't consistent with other similar traits such "Skill Versality" for the Half Elf as per issue #187.

### DIFF
--- a/src/5e-SRD-Traits.json
+++ b/src/5e-SRD-Traits.json
@@ -104,13 +104,7 @@
         "url": "/api/races/dwarf"
       }
     ],
-    "subraces": [
-      {
-        "index": "hill-dwarf",
-        "name": "Hill Dwarf",
-        "url": "/api/subraces/hill-dwarf"
-      }
-    ],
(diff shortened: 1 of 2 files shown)


src/5e-SRD-Traits.json
```diff
@@ -95,6 +95,28 @@
...
+        "url": "/api/races/dwarf"
+      }
+    ],
+    "subraces": [
+      {
+        "index": "hill-dwarf",
+        "name": "Hill Dwarf",
+        "url": "/api/subraces/hill-dwarf"
+      }
+    ],
```
bagelbits: If this is given with dwarf, we don't really need to add it with Hill Dwarf. Unless you're granted it a second time with Hill Dwarf.

bagelbits:
```suggestion
    "subraces": [],
```

bagelbits (approved):
Awesome! Thank you!

PR #284, opened 2020-09-30
[WIP] first batch of features data update

Freedzio:
I took the liberty of deleting the template.

The structures (besides the ones discussed with @fergcb) are experimental, so I expect a lot of comments :P

### DIFF
--- a/src/5e-SRD-Features.json
+++ b/src/5e-SRD-Features.json
@@ -24,7 +24,6 @@
         {
           "name": "Rage",
           "action_type": "bonus",
-          "duration": "10 turns",
           "prerequisites": [
             {
               "type": "boolean",
@@ -33,6 +32,7 @@
           ],
           "effects": [
             {
+              "duration": "10 turns",
               "type": "advantage",
               "on": {
                 "trigger": "check",
@@ -47,6 +47,7 @@
               }
             },
             {
+              "duration": "10 turns",
               "type": "advantage",
               "on": {
                 "trigger": "save",
@@ -62,6 +63,7 @@
             },
             {
               "type": "bonus",
+              "duration": "10 turns",
               "amount": 2,
               "to": "damage",
               "on": {
@@ -79,6 +81,7 @@
             },
             {
               "type": "resistance",
+              "duration": "10 turns",
               "to": "damage",
               "target": "self",
               "source": "any",
@@ -92,6 +95,7 @@
             },
             {
               "type": "resistance",
+              "duration": "10 turns",
               "to": "damage",
               "target": "self",
               "source": "any",
@@ -105,6 +109,7 @@
             },
             {
               "type": "resistance",
+              "duration": "10 turns",
               "to": "damage",
               "target": "self",
               "source": "any",
@@ -117,7 +122,11 @@
               }
             },
             {
-              "spellcasting": false
+              "type": "action_restriction",
+              "deny_actions": [
+                "spellcasting"
+              ],
+              "duration": "10 turns"
             }
           ]
         }
@@ -177,7 +186,7 @@
       "actions": [
         {
           "name": "Reckless Attack",
-          "desc": "When you make your first attack on your turn...",
+          "desc": "When you make your first attack on your turn, you can decide to attack recklessly. Doing so gives you advantage on melee weapon attack rolls using Strength during this turn, but attack rolls against you have advantage until your next turn.",
           "action_type": "attack",
           "prerequisites": [
             {
@@ -335,12 +344,15 @@
       "actions": [
         {
           "name": "Frenzy",
-          "desc": "When you enter rage, you can go into frenzy instead",
+          "desc": [
+            "You go into a frenzy when you rage. For the duration of your rage you can make a single melee weapon attack as a bonus action on each of your turns after this one. When your rage ends, you suffer one level of exhaustion (as described in appendix PH-A)",
+            " If you do so, for the duration of your rage you can make a single melee weapon attack as a bonus action on each of your turns after this one. When your rage ends, you suffer one level of exhaustion (as described in appendix PH-A)."
+          ],
           "action_type": "bonus",
-          "duration": "10 turns",
           "effects": [
             {
               "type": "bonus_action",
+              "duration": "10 turns",
               "bonus_action_type": "attack",
               "where": {
                 "attack_type": "melee"
@@ -348,6 +360,7 @@
             },
             {
               "type": "condition",
+              "duration": "10 turns",
               "condition_type": {
                 "index": "exhaustion",
                 "name": "Exhaustion",
--- a/src/5e-SRD-Features.json
+++ b/src/5e-SRD-Features.json
@@ -40,8 +40,7 @@
         "values": [
           {
             "type": "integer",
-            "value": 10,
-            "desc": "Base value for calculation"
+            "value": 10
           },
           {
             "type": "integer",
@@ -72,31 +71,35 @@
       "Starting at 2nd level, you can throw aside all concern for defense to attack with fierce desperation. When you make your first attack on your turn, you can decide to attack recklessly. Doing so gives you advantage on melee weapon attack rolls using Strength during this turn, but attack rolls against you have advantage until your next turn."
     ],
     "feature_specific": {
-      "advantages": [
+      "action": [
         {
-          "on": [
-            {
-              "action": "attack",
-              "attack_type": "melee",
-              "attack_modifier": "str",
-              "desc": "Advantage applies to melee attacks with STR modifier only"
-            }
-          ],
-          "duration": "2 turns",
-          "prerequisites": [
-            {
-              "type": "integer",
-              "from": "attack",
-              "desc": "Advantage applies to first attack only"
-            }
-          ]
+          "type": "attack",
+          "attack_type": "melee",
+          "attack_modifier": "str",
+          "desc": "Advantage applies to melee attacks with STR modifier only"
         }
       ],
-      "downsides": [
+      "prerequisites": [
+        {
+          "type": "equals",
+          "description": "attack is the first of the turn",
+          "value": {
+            "type": "integer",
+            "value": "turns.attacks_made",
+            "desc": "attacks made so far this turn"
+          },
+          "target": {
+            "type": "integer",
+            "value": 0
+          }
+        }
+      ],
+      "debuffs": [
         {
           "type": "advantage",
           "for": "enemy",
           "on": "attack",
+          "duration": "2 turns",
           "desc": "Attack rolls against you have advantage until the end of your next turn"
         }
       ]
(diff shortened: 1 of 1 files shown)


src/5e-SRD-Features.json
```diff
@@ -18,7 +18,8 @@
       "Your rage lasts for 1 minute. It ends early if you are knocked Unconscious or if Your Turn ends and you haven't attacked a hostile creature since your last turn or taken damage since then. You can also end your rage on Your Turn as a Bonus Action.",
       "Once you have raged the maximum number of times for your barbarian level, you must finish a Long Rest before you can rage again. You may rage 2 times at 1st level, 3 at 3rd, 4 at 6th, 5 at 12th, and 6 at 17th."
     ],
-    "url": "/api/features/rage"
+    "url": "/api/features/rage",
+    "Feature_specific": {}
```
fergcb: Typo here - `"Feature_specific"` shouldn't be capitalised.

src/5e-SRD-Features.json
```diff
@@ -33,6 +34,28 @@
...
       "While you are not wearing any armor, your Armor Class equals 10 + your Dexterity modifier + your Constitution modifier. You can use a shield and still gain this benefit."
     ],
+    "feature_specific": {
+      "armor_class": {
+        "type": "sum",
+        "values": [
+          {
+            "type": "integer",
+            "value": 10,
+            "desc": "Base value for calculation"
```
fergcb: I feel like straight up "base" values for calculations probably don't need descriptions?

src/5e-SRD-Features.json
```diff
@@ -63,6 +116,32 @@
     "desc": [
       "At 2nd level, you gain an uncanny sense of when things nearby aren't as they should be, giving you an edge when you dodge away from danger. You have advantage on Dexterity saving throws against effects that you can see, such as traps and spells. To gain this benefit, you can't be blinded, deafened, or incapacitated."
     ],
+    "feature_specific": {
+      "advantages": [
+        {
+          "on": [
+            {
+              "action": "save",
+              "save_type": "dex"
```
fergcb: Might be nice to have an APIReference to the actual ability here?

src/5e-SRD-Features.json
```diff
@@ -48,6 +71,36 @@
...
+              "attack_type": "melee",
+              "attack_modifier": "str",
+              "desc": "Advantage applies to melee attacks with STR modifier only"
+            }
+          ],
+          "duration": "2 turns",
+          "prerequisites": [
+            {
+              "type": "integer",
+              "from": "attack",
```
fergcb: Maybe something like `turn.attacks_made` or `round.attack_number`, or something like that?
Also, ought all prerequisites be booleans (or operations resulting in booleans)? Something like the following?

```json
{
    "type": "equals",
    "description": "attack is the first of the turn",
    "value": {
        "type": "integer",
        "value": "turns.attacks_made",
        "desc": "attacks made so far this turn"
     },
    "target": {
        "type": "integer",
        "value": 0
    }
}
```

src/5e-SRD-Features.json
```diff
@@ -48,6 +71,36 @@
...
+              "desc": "Advantage applies to first attack only"
+            }
+          ]
+        }
+      ],
+      "downsides": [
```
fergcb: Not a big fan of the name "downsides". What about "debuffs"?

src/5e-SRD-Features.json
```diff
@@ -48,6 +71,36 @@
     "desc": [
       "Starting at 2nd level, you can throw aside all concern for defense to attack with fierce desperation. When you make your first attack on your turn, you can decide to attack recklessly. Doing so gives you advantage on melee weapon attack rolls using Strength during this turn, but attack rolls against you have advantage until your next turn."
     ],
+    "feature_specific": {
+      "advantages": [
```
fergcb: Is there a way we could wrap this in some kind of "action" structure? So the client knows it can be added to the character's possible list of actions. The action would grant this advantage. Just a thought, maybe one for a future PR though.

src/5e-SRD-Features.json
```diff
@@ -63,6 +116,32 @@
...
+          "on": [
+            {
+              "action": "save",
+              "save_type": "dex"
+            }
+          ],
+          "prerequisites": [
+            {
+              "type": "boolean",
+              "from": "incapacitated"
```
fergcb: Maybe `conditions.incapacitated`? Same for the other conditions below.

src/5e-SRD-Features.json
```diff
@@ -97,6 +189,24 @@
...
+          "from": "raging"
+        }
+      ],
+      "bonus_action": {
+        "action": "attack",
+        "amount": 1
+      },
+      "downsides": [
+        {
+          "type": "exhaustion",
```
fergcb: What about something more like the following?

```json
{
  "type": "condition",
  "condition": "exhaustion",
  "level": 1
}
```

src/5e-SRD-Features.json
```diff
@@ -127,6 +237,9 @@
     "desc": [
       "Beginning at 5th level, you can attack twice, instead of once, whenever you take the Attack action on your turn."
     ],
+    "feature_specific":{
+
```
fergcb: Erroneous white-space

src/5e-SRD-Features.json
```diff
@@ -142,6 +255,27 @@
...
+          },
+          {
+            "type": "integer",
+            "value": 10
+          }
+        ],
+        "prerequisites": [
+          {
+            "type": "boolean",
+            "from": "heavy armor"
```
fergcb: Instead of `"heavy armor"`, what about something like `equipped.heavy_armor`?

src/5e-SRD-Features.json
```diff
@@ -8774,4 +8978,4 @@
     ],
     "url": "/api/features/signature-spell"
   }
-]
+]
```
fergcb, line 9156: poor old newline :( my IDE does this too! Probs best to keep it though.

fergcb (changes requested):

src/5e-SRD-Features.json
Freedzio: hmm, you are right. I got too carried away with describing things

src/5e-SRD-Features.json
Freedzio: I like that approach

src/5e-SRD-Features.json
Freedzio: maybe an APIreference?

fergcb: Yeah, an APIReference to the condition would be ideal, actually!

src/5e-SRD-Features.json
Freedzio: I'd go with `equipped.armor.heavy`, since there is no `heavy_armor` label when it comes to equipment. Item can can have category `armor`, and then a prop specifying the heaviness - `armor_category`

fergcb: Yeah, that works

src/5e-SRD-Features.json
```diff
@@ -72,31 +71,35 @@
...
-              "from": "attack",
-              "desc": "Advantage applies to first attack only"
-            }
-          ]
+          "type": "attack",
+          "attack_type": "melee",
+          "attack_modifier": "str",
+          "desc": "Advantage applies to melee attacks with STR modifier only"
         }
       ],
```
Freedzio: I feel like someting is missing here tho. The `action` wrapper

fergcb: This isn't quite what I was describing.

I meant that there should be some action object, describing the (bonus) action the player can take to enter a rage.

*Within* that action object, we'd have the advantages and debuffs as described in your original PR. Simply wrapping the effects of Rage in the action that causes them.

I'll type up an example tomorrow.

src/5e-SRD-Features.json
```diff
@@ -19,7 +19,110 @@
       "Once you have raged the maximum number of times for your barbarian level, you must finish a Long Rest before you can rage again. You may rage 2 times at 1st level, 3 at 3rd, 4 at 6th, 5 at 12th, and 6 at 17th."
     ],
     "url": "/api/features/rage",
-    "feature_specific": {}
+    "feature_specific": {
+      "actions": [
+        {
+          "name": "Rage",
+          "action_type": "bonus",
+          "duration": "10 turns",
```
fergcb: Actions don't really have durations, effects do. I think having a `duration` attribute on each effect makes more sense than the action itself having a duration.

src/5e-SRD-Features.json
```diff
@@ -19,7 +19,110 @@
...
+              "where": {
+                "damage_type": {
+                  "index": "slashing",
+                  "name": "Slashing",
+                  "url": "/api/damage-types/slashing"
+                }
+              }
+            },
+            {
+              "spellcasting": false
```
fergcb: What about
```json
{
  "type": "action_restriction",
  "deny_actions": ["spell"],
  "duration": "10 turns"
}
```

src/5e-SRD-Features.json
```diff
@@ -71,36 +174,56 @@
...
-      "debuffs": [
+      "actions": [
         {
-          "type": "advantage",
-          "for": "enemy",
-          "on": "attack",
-          "duration": "2 turns",
-          "desc": "Attack rolls against you have advantage until the end of your next turn"
+          "name": "Reckless Attack",
+          "desc": "When you make your first attack on your turn...",
```
fergcb: This should be the full description of the action. From the SRD:

> When	you	make	your	first	attack	on	your	turn,	you	can	
decide	to	attack	recklessly.	Doing	so	gives	you	
advantage	on	melee	weapon	attack	rolls	using	
Strength	during	this	turn,	but	attack	rolls	against	
you	have	advantage	until	your	next	turn.

src/5e-SRD-Features.json
```diff
@@ -206,22 +332,30 @@
...
-      ],
-      "debuffs": [
+      "actions": [
         {
-          "index": "exhaustion",
-          "name": "Exhaustion",
-          "url": "/api/conditions/exhaustion",
-          "level": 1
+          "name": "Frenzy",
+          "desc": "When you enter rage, you can go into frenzy instead",
```
fergcb: Use the description from the SRD:

> You go	into	a	frenzy	when	you	rage.	For	
the	duration	of	your	rage	you	can	make	a	single	
melee	weapon	attack	as	a	bonus	action	on	each	of	
your	turns	after	this	one.	When	your	rage	ends,	you	
suffer	one	level	of	exhaustion	(as	described	in	
appendix	PH-A).

src/5e-SRD-Features.json
```diff
@@ -206,22 +332,30 @@
...
+          "action_type": "bonus",
+          "duration": "10 turns",
```
fergcb: Same thing with `duration` as above.

src/5e-SRD-Features.json
Freedzio: ok, with `spellcasting` instead of `spells`. sounds rightier to me

src/5e-SRD-Features.json
Freedzio: ok. I've also split it into an array

src/5e-SRD-Features.json
```diff
@@ -335,19 +344,23 @@
       "actions": [
         {
           "name": "Frenzy",
-          "desc": "When you enter rage, you can go into frenzy instead",
+          "desc": [
+            "You go into a frenzy when you rage. For the duration of your rage you can make a single melee weapon attack as a bonus action on each of your turns after this one. When your rage ends, you suffer one level of exhaustion (as described in appendix PH-A)",
+            " If you do so, for the duration of your rage you can make a single melee weapon attack as a bonus action on each of your turns after this one. When your rage ends, you suffer one level of exhaustion (as described in appendix PH-A)."
```
fergcb, line 349: Seems like the second item here is a duplicate.

bagelbits:
@Freedzio @fergcb Y'all still working on this?

fergcb:
> @Freedzio @fergcb Y'all still working on this?

I haven't heard back from @Freedzio in regards to some of my comments on what he's got so far.

bagelbits:
Closing for now as it hasn't been touched for a while and we're migrating to a monorepo

PR #293, opened 2020-10-06
Add proficiency information to traits, proficiencies granted by traits to half elf and rock gnome

Ajhaa:
## What does this do?
Adds proficiencies that come from traits to half elf and rock gnome

## How was it tested?
Manually verified

#187

### DIFF
--- a/src/5e-SRD-Races.json
+++ b/src/5e-SRD-Races.json
@@ -598,7 +598,102 @@
     "size": "Medium",
     "size_description": "Half-elves are about the same size as humans, ranging from 5 to 6 feet tall. Your size is Medium.",
     "starting_proficiencies": [],
-    "starting_proficiency_options": {},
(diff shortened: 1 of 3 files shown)


bagelbits:
I feel like this is half the issue. Since they should be in the traits as well.

bagelbits:
Okay. This looks great.
One last thing though, can you open a PR with https://github.com/bagelbits/5e-srd-api to update the docs. You should only need to change [this file.](https://github.com/bagelbits/5e-srd-api/blob/master/src/views/partials/doc-resource-traits.ejs)

I will happily merge this in once that is done.

bagelbits:
Nevermind, I handled it here:
https://github.com/bagelbits/5e-srd-api/pull/116

bagelbits (approved):
LGTM

Issue #299, opened 2020-10-12
Monster Data should include XP


Redmega:
As per bagelbits/5e-srd-api#98 XP is not strictly a simple lookup table -- XP varies for 0CR creatures based on whether or not they have any attack actions available to them.

![](https://camo.githubusercontent.com/a8a0c76a2854f46b28056024467a14a508d7b8b4/68747470733a2f2f6d656469612e646973636f72646170702e6e65742f6174746163686d656e74732f3635363534373636373630313635333739302f3735383435373931343837353936393537362f53637265656e73686f745f32303230303932335f3233343035352e6a70673f77696474683d323330266865696768743d323536)

txtsd:
Here's a list of 0CR monsters:
```
Awakened Shrub
Baboon
Badger
Bat
Cat
Commoner
Crab
Deer
Eagle
Frog
Giant Fire Beetle
Goat
Hawk
Homunculus
Hyena
Jackal
Lemure
Lizard
Octopus
Owl
Quipper
Rat
Raven
Scorpion
Sea Horse
Shrieker
Spider
Vulture
Weasel
```

Only these don't have a damaging action:
```
Frog
Sea Horse
Shrieker
```
Shrieker has an action called `Shriek` that says `When bright light or a creature is within 30 feet of the shrieker, it emits a shriek audible within 300 feet of it. The shrieker continues to shriek until the disturbance moves out of range and for 1d4 of the shrieker's turns afterward` so it's not an attack.

So should I do `XP: 10` for all except Frog, Sea Horse and Shrieker?

Redmega:
According to the SRD shriekers are worth 10xp. The only 0xp monsters are Frog, Sea Horse, and Avatar of Death (A notable exception, this one has an attack but is CR-0 with 0xp as per SRD. I wonder if that's an errata? Or if it's accurate because the only way it appears is if summoned by a Deck of Many Things)

ragnork2:
Oh i should mention that ive already been working on this

txtsd:
> According to the SRD shriekers are worth 10xp.

You're right.


> Actually, the Avatar of Death doesn't seem to be in the ~SRD~ API. Wonder if that was on purpose (Since it can only be summoned by a Deck of Many Things) or if it's just coincidental and needs to be added in different PR.

It's missing because it's listed under Magic Items, not Monsters.

Redmega:
Oh, good catch! Didn't notice -- Just did a search for (0 XP).

[...]

PR #301, opened 2020-10-13
Add Avatar of Death and Giant Fly

txtsd:
## What does this do?
Adds the Avatar of Death and Giant Fly. They were missing because they were listed under Magic Items.

## How was it tested?
Added manually.

No, but mentioned here #299 

No

## More
Things that are different from the norm:
For `Avatar of Death`: 
`"hit_points": "half the hit point maximum of its summoner"`
 `hit_dice` missing
`"challenge_rating": "—",` like in the SRD. Should I make it 0?
For `Giant Fly`:
`challenge_rating` and `actions` and `special_abilties` are missing

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -6850,6 +6850,102 @@
     "url": "/api/monsters/assassin"
   },
   {
+    "index": "avatar-of-death",
+    "name": "Avatar of Death",
+    "size": "Medium",
+    "type": "undead",
+    "subtype": null,
+    "alignment": "neutral evil",
+    "armor_class": 20,
+    "hit_points": "half the hit point maximum of its summoner",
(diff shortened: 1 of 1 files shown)


bagelbits:
For `Giant Fly`, CR is most likely 0. And `actions` and `special_abilities` are missing because it has none.

For `Avatar of Death`, wow. That's.... that's really weird. I almost want to not even put that in here because of how weird it is and leave it in Magic Items.

Redmega:
The Giant Fly only comes from `Figurine of Wondrous Power` (A magical item) and is also lacking a CR and XP entry in the SRD

bagelbits:
I'm honestly okay if we just add neither?

txtsd:
Fair enough. I'll leave this here anyway in case we change our minds.

PR #305, opened 2020-10-16
Add xp to monsters

ragnork2:
## What does this do?
Adds an xp field to monsters

## How was it tested?
VScode linter

Issue #299 

I did not, not sure if i needed to, lmk if i do
random photo - update me


src/5e-SRD-Monsters.json
```diff
@@ -40348,4 +40670,4 @@
     ],
     "url": "/api/monsters/zombie"
   }
-]
+]
```
bagelbits, line 40673:
```suggestion
]

```

bagelbits (approved):
Yeah! Nice!

src/5e-SRD-Monsters.json
ragnork2: I have no idea why VScode decided to do that lol

bagelbits: It's probably a setting somewhere?

fergcb: @ragnork2 The setting in VS code is called "insert final newline". You should be able to search for it in the settings menu.

ragnork2: ah kk thanks!

ogregoire:
Since the CR and the XP are strongly related, it might have been a good idea to have made a structure of the two, like:

```
{
  challenge: {
    rating: "0",
    xp: 0
  }
}
```

bagelbits:
@ogregoire You raise a valid point.

fergcb:
I can't say I'm a fan of that idea.

It complicates the structure with no real benefit for the client, as far as I can see; this structure isn't exactly reusable, and having the properties belong directly to the monster isn't exactly confusing.

It also goes against the pattern of query parameters having the same names as the attribute they correspond to in the document.

If we *were* to make this change, I don't think `challenge` is a suitable name. Sure, XP comes from CR, but it is not *part* of the challenge rating itself. That said, I can't think of anything more appropriate...

ogregoire:
Yes, "Challenge" is the name. See SRD p. 258.

fergcb:
"Challenge" refers to the challenge rating. The XP in brackets is not a component of the "challenge", but a result of it. 

I read "**Challenge** 2 (450xp)" not as `challenge: { rating: 2, xp:450 }`, but as `challenge: 2, (xp: 450)`.

"Challenge" is an appropriate label for the challenge rating, but I don't feel that label reflects the fact that XP is included in that property.

PR #313, opened 2020-10-24
references index name, not position

sdrewdle:
## What does this do?
Removes references to the position of the ability scores in their json, instead references their names.

## How was it tested?
ran locally

no

n/a

random photo - update me


bagelbits (approved):
Good catch!

PR #317, opened 2020-11-03
Update 5e-SRD-Magic-Items.json

johnsenner:
Cheers!

## What does this do?
Adds url for the crab magic item

## How was it tested?
I checked that url, it works

no, just found missing data while parsing

n/a

random photo - update me


bagelbits (approved):
Good catch!

PR #323, opened 2020-11-24
Fixing inconsistencies

tognee:
## What does this do?
Adds missing blank objects where they should be

## How was it tested?
Parsed with a JSON parser and tested with a python script

### DIFF
--- a/src/5e-SRD-Features.json
+++ b/src/5e-SRD-Features.json
@@ -6,9 +6,9 @@
       "name": "Barbarian",
       "url": "/api/classes/barbarian"
     },
-    "subclass": {},
     "name": "Rage",
     "level": 1,
+    "prerequisites": [],
     "desc": [
       "In battle, you fight with primal ferocity. On your turn, you can enter a rage as a bonus action. While raging, you gain the following benefits if you aren't wearing heavy armor:",
       "- You have advantage on Strength checks and Strength saving throws.",
@@ -27,9 +27,9 @@
       "name": "Barbarian",
       "url": "/api/classes/barbarian"
     },
-    "subclass": {},
     "name": "Unarmored Defense",
     "level": 1,
+    "prerequisites": [],
     "desc": [
       "While you are not wearing any armor, your Armor Class equals 10 + your Dexterity modifier + your Constitution modifier. You can use a shield and still gain this benefit."
     ],
@@ -42,9 +42,9 @@
       "name": "Barbarian",
       "url": "/api/classes/barbarian"
     },
-    "subclass": {},
     "name": "Reckless Attack",
     "level": 2,
+    "prerequisites": [],
     "desc": [
       "Starting at 2nd level, you can throw aside all concern for defense to attack with fierce desperation. When you make your first attack on your turn, you can decide to attack recklessly. Doing so gives you advantage on melee weapon attack rolls using Strength during this turn, but attack rolls against you have advantage until your next turn."
     ],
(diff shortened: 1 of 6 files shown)


fergcb:
Starting equipment fix looks good.

Can't say I'm a big fan of all the empty objects, though. I'd personally prefer these fields to be optional. With the empty objects there, the presence of the object suggests the value exists, but on inspection, the value would be invalid.

Perhaps splitting these two unrelated proposals into different PRs would be a good idea, so the fix can go through while we discuss the empty objects proposal?

tognee:
Moved the paladin fix to #324 

I've added the blank objects as there were already blank objects in just some properties, but not all.
This was brought to my attention by my python script.

The solution should be adding the empty objects (as this pr does) or remove them all, while keeping only the ones that matter

fergcb:
My vote would be to axe the empties, for sure. An empty object doesn't sufficiently represent the absence of an optional field, imo.

tognee:
Take a look at all the files, here are my notes:
Didn't check Monsters and Levels empty arrays, as I can't rn
Let me know your thoughts

Empty arrays

Ability-Scores
"con" -> "skills"
It should stay as it has no skills, unlike the other abilities

Equipment
"flail", "morningstar" and "war-pick" have "properties" as an empty Array
It should stay as all the other equipments have properties

Features
11 Features have "prerequisites" as an Empty Array
It should be removed as other Features without prerequisites don't have an empty Array

Monsters
1122 empty Arrays

Levels
334 empty Arrays

Proficiencies
All empty Arrays refer to classes and races
Should stay

Races
"starting_proficiencies", "traits" and "subraces" have empty Arrays
IMO should stay if empty

Spells
All refer to "subclasses"
Should stay

StartingEquipment
empty starting_equipment in "fighter"
Should stay

Subraces
"starting_proficiencies" and "languages" have empty Arrays
Should stay

Traits
"races", "subraces" and "proficiencies"
Should stay

Empty objects

Features
"subclasses" have empty objects
Should stay

Levels
"subclass" should stay
empty "spellcasting" should be removed on all classes and subclasses that can't use spells
Maybe removing from subclasses as well?

Races
"ability_bonus_options", "starting_proficiency_options", "language_options", "trait_options"
IMO should stay if empty

Spells
"heal_at_slot_level" in "greater-restoration"
Not sure

Subraces
"ability_bonus_options", "starting_proficiency_options", "language_options", "racial_trait_options"
IMO should stay if empty

fergcb:
Empty arrays should stay, absolutely. I only recommended that empty *objects* should be cut.

I don't think *any* of the attributes you listed with object values should stay if empty. Ones ending in `_option`, for example, represent choices. If there is no choice to be made, the attribute probably shouldn't be there, rather than being an empty object (which looks at first glance like something should be there, but doesn't actually follow the schema of an Choice object).

package-lock.json
```diff
@@ -0,0 +1,6520 @@
+{
```
bagelbits: Currently, we're using yarn for this repo, which is why there is no `package-lock.json`. Can you remove this for now?

bagelbits:
I think I still see some empty arrays that you removed which should be put back?

tognee:
I removed "prerequisites" in Features as not all had that array
Should I put back only the ones I removed or also add the missing ones?

ogregoire:
This seems good. Given the various forths and backs, I suggest we squash the commits.

bagelbits:
I believe you'll want to rebase.

PR #324, opened 2020-11-24
Fix paladin wrapping

tognee:
## What does this do?
Fixes inconsistent structure of the Paladin's StartingEquipment (simple-weapons wasn't wrapped by equipment_category)

## How was it tested?
Parsed with a JSON parser and tested with a python script


bagelbits (approved):
LGTM!

PR #332, opened 2020-12-20
Fixes dice notation for a couple spells

guillianbalisi:
## What does this do?

Fixes dice notation for Phantasmal Killer and Weird, there was an extra space between the number of dice and which dice to use.

There are other cases where the dice notation doesn't include "1" when rolling a single die, e.g. Bane: "target must roll a d4 and subtract the number". Does it make sense to turn this into "target must roll 1d4 and subtract the number"?

## How was it tested?

CI ✅ 

No, just found this while looking at the spells

n/a

random photo - update me


bagelbits (approved):
LGTM!

PR #338, opened 2021-01-14
monsters/ankheg updated damage type

brittonhayes:
## What does this do?
Updated the damage type for the Ankheg to be of type `acid` rather than of type `bludgeoning`. Also updates the associated index route listed in the Ankheg's damage type.

## How was it tested?
NA

NA

NA


bagelbits (approved):
LGTM!

Issue #251, opened 2020-08-31
Merge Spellcasting, StartingEquipment, and Class Levels into Class objects.


fergcb:
Currently, the 1:1 relationships of Class->Spellcasting and Class->StartingEquipment exist in the schema as URL strings, rather than the data simply being embedded in the Class object. This only serves to reduce the size of the Class endpoint response, but complicates the digestion of the data rather a lot. Having separate endpoints for Spellcasting and StartingEquipment also creates difficulty maintaining consistency in the structure of internal references; where all other references are suited to using the APIResource structure, these are simply incompatible due to the shape of the referenced data.

As for Class Levels, the size cost of including *all* Level data in the class resource is much greater than for Spellcasting or StartingEquipment. It would perhaps be more suitable, however, to provide references to class levels as an array of APIResources, rather than the current method of supplying a URL that returns such a list. This would again simplify the digestion of the API data greatly, including by reducing the number of API calls that have to be made. Additionally, leaving the Levels endpoint intact allows levels to be queried separately from Classes, if desired.

**In summary:**
- Merge Spellcasting and StartingEquipment data directly into Class objects, get rid of corresponding API endpoints
- Provide Class Levels as an array of APIResources in the Class object, keeping Levels endpoint

All of this can be applied to Subclasses, also.

Another one to discuss is class Spell lists. Move them into Spellcasting, or directly onto the Class object, as an array of APIResources, like Levels?

bagelbits:
Spellcasting is now merged.

bagelbits:
> All of this can be applied to Subclasses, also.

I would leave subclasses as is. Mainly because you can have more than one subclass, and it would break things for anyone that adds any additional subclass for their own DB. Or if more subclasses ever get released for OGL.

fergcb:
@bagelbits that's not what I meant by that comment. I don't mean that subclasses should be merged into classes like levels, starting equipment, etc.

I mean that subclass levels could be merged into subclass documents as arrays (and any other 1-1 relationships), in the same way I proposed for classes.

bagelbits:
Oh! My mistake! That makes a lot of sense!

bagelbits:
Closing as mostly done: `spellcasting` and `starting_equipment` are now embedded in class objects. `class_levels` and `spells` remaining as URLs is an intentional design choice.

PR #342, opened 2021-02-05
Feature/backgrounds

Arnoldave:
Fine, I'll do it myself

## What does this do?
This adds the one SRD background, Acolyte, to the database. It's done so in a way that implies others exist (i.e., in a list). Also included is it's feature, Shelter of the Faithful, stored alongside class features, and it's starting equipment, stored alongside starting equipment for classes.

## How was it tested?
I've built, and ran the db as a docker image, and checked to make sure the data I added is in there. I attempted to run the unit tests, however it threw this error: `npm ERR! missing script: test:unit`. Happy to go back and run it if there's an easy answer.

Not to my knowledge, though I don't doubt you've been asked about this a lot.

Not yet. Will update the docs as a part of my changes to the API (which I've begun work on). I'm too shit at Docker to connect to my local branch of this, so was planning to throw up this PR partly as a cry for help.


src/5e-SRD-Backgrounds.json
```diff
@@ -0,0 +1,147 @@
...
+                }
+            ]
+        },
+        "starting_equipment": "/api/starting-equipment/acolyte",
+        "feature": {
+            "index": "shelter-of-the-faithful",
+            "name": "Shelter of the Faithful",
+            "url": "/api/features/shelter-of-the-faithful"
+        },
+        "suggested_characteristics": {
```
bagelbits: You can probably make use of the choice structure for these too.

Arnoldave: I could do. Mind you, AFAIK there's no specific limit to how many of each you can choose, so I may not be able to assign the `choose` value.

fergcb: In the "Customising a Background" section in the SRD, it says the following:
> Finally, choose two personality traits, one ideal, one bond, and one flaw

fergcb: But even if we don't use Choices for these, we should probably consider using arrays, rather than numerical keys like this.

src/5e-SRD-StartingEquipment.json
```diff
@@ -1337,5 +1337,46 @@
...
+          "equipment_category": {
+            "index": "holy-symbols",
+            "name": "Holy Symbols",
+            "url": "/api/equipment-categories/holy-symbols"
+          }
+        }
+      }
+
+    ],
+    "url": "/api/starting-equipment/wizard"
```
fergcb: Looks like you missed a spot! The URL should be `/api/starting-equipment/acolyte`.

src/5e-SRD-Backgrounds.json
Arnoldave: This was changed to use those exact values you mentioned. The numbers represent die values you would roll to get these characteristics if you wish to roll for them (as detailed in the SRD). The updated structure reflects this a bit better.

fergcb (changes requested):
A couple of little comments. Otherwise, LGTM!

src/tests/5e-SRD-Alignments.json
```diff
@@ -0,0 +1,64 @@
+[
+    {
+        "index": "lg",
```
fergcb: I can't say I'm a big fan of using the abbreviation for the `index` and `name`, here. I feel like supplying the abbreviation as a separate property might be more suitable?

```json
{
  "index": "lawful-good",
   "name": "Lawful Good",
   "abbreviation": "LG",
   "desc": "Lawful good (LG) creatures can be counted on to do the right thing as expected by society. Gold dragons, paladins, and most dwarves are lawful good.",
   "url": "/api/alignments/lawful-good"
}
```

Arnoldave:
I've made the suggested change to alignments.

src/5e-SRD-Suggested-Characteristics.json
```diff
@@ -0,0 +1,270 @@
...
+            "index": "acolyte",
+            "name": "Acolyte",
+            "url": "api/backgrounds/acolyte"
+        },
+        "personality_traits": {
+            "choose": 2,
+            "type": "personality_traits",
+            "from": [
+                {
+                    "die_value": 1,
```
bagelbits: The die value seems redundant since they're already in an array.

src/5e-SRD-Alignments.json
```diff
@@ -0,0 +1,65 @@
+[
```
bagelbits, line 1: 👏🏼 👏🏼 👏🏼 👏🏼 👏🏼

bagelbits:
I have a vague concern about combining the background starting equipment/features with classes. However, not enough to block on or have a good suggestion for. It's good enough for now unless someone has a thought.

src/5e-SRD-Backgrounds.json
```diff
@@ -0,0 +1,112 @@
...
+        "starting_equipment": "/api/starting-equipment/acolyte",
+        "feature": {
+            "index": "shelter-of-the-faithful",
+            "name": "Shelter of the Faithful",
+            "url": "/api/features/shelter-of-the-faithful"
+        },
+        "suggested_characteristics": "api/suggested-characteristics/acolyte",
+        "url": "api/backgrounds/acolyte"
+    }
+]
```
bagelbits: Nit pick: Add new line at the end of the files that don't have it. If you use an IDE or code editor, there should be a setting to do this.

bagelbits (approved):
LGTM!

src/5e-SRD-Suggested-Characteristics.json
```diff
@@ -0,0 +1,204 @@
+[
```
bagelbits: Since it's a 1:1 relationship, is there any reason we shouldn't just combine Suggested Characteristics with Backgrounds?

src/5e-SRD-StartingEquipment.json
```diff
@@ -1337,5 +1337,45 @@
       }
     ],
     "url": "/api/starting-equipment/wizard"
+  },
```
bagelbits: I just remembered that we have an [open issue](https://github.com/5e-bits/5e-database/issues/251) from @fergcb to combine StartingEquipment into Classes. So this may also be better to just be folded into Backgrounds since this is also a 1:1 relationship.

bagelbits: Which finally got my butt in gear to put together [this PR](https://github.com/5e-bits/5e-database/pull/343).

bagelbits (commented):
Thanks! I like that a lot better. It should cleanup your API PR a bit too.

src/5e-SRD-Features.json
```diff
@@ -8812,5 +8812,19 @@
...
+      "index": "acolyte",
+      "name": "Acolyte",
+      "url": "/api/backgrounds/acolyte"
+    },
+    "name": "Shelter of the Faithful",
+    "desc": [
+      "As an acolyte, you command the respect of those who share your faith, and you can perform the religious ceremonies of your deity. You and your adventuring companions can expect to receive free healing and care at a temple, shrine, or other established presence of your faith, though you must provide any material components needed for spells. Those who share your religion will support you (but only you) at a modest lifestyle.",
+      "You might also have ties to a specific temple dedicated to your chosen deity or pantheon, and you have a residence there. This could be the temple where you used to serve, if you remain on good terms with it, or a temple where you have found a new home. While near your temple, you can call upon the priests for assistance, provided the assistance you ask for is not hazardous and you remain in good standing with your temple."
+    ],
+    "url": "/api/features/shelter-of-the-faithful"
```
bagelbits: Per conversations in Discord, since backgrounds always have one feature and that feature is unique to the background, it makes more sense to bundle this into repo. And it also answers the thing that was tickling me about this addition.

bagelbits (approved):
Golden! Sorry for all the back and forth on this one.

PR #343, opened 2021-02-12
Merge StartingEquipment into Class

bagelbits:
## What does this do?
This is the first of two steps. I'm planning to merge the StartingEquipment into Class on this step. The second step will be cleaning up and removing the StartingEquipment file.

## How was it tested?
CI

This is related to #251 

Not yet. But I will put together a PR to mirror these changes.

image

### DIFF
--- a/src/5e-SRD-Classes.json
+++ b/src/5e-SRD-Classes.json
@@ -80,7 +80,80 @@
         "url": "/api/ability-scores/con"
       }
     ],
-    "starting_equipment": "/api/starting-equipment/barbarian",
+    "starting_equipment": [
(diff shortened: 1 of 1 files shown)


bagelbits:
I decided that it might make more sense to do this in a two step process because we can and it will be nice for our users. I'm going to do the merge first and setup the API. Then I'll make an announcement in the discord that StartingEquipment is going away in about a week and the new endpoint for it is already available. And then I can do the fun of cleanup.

fergcb (approved):
LGTM! Doing this in two stages makes good sense.

PR #353, opened 2021-03-16
Correct Aboleth tentacle attack

youknowme786:
Co-authored-by: ZacChristensen31 <zac.d.christensen@gmail.com>

## What does this do?
This fixes a bug with the Aboleth's tentacle attack.

## How was it tested?
Compared to 5E SRD, as well as the description of the attack in the repo.

No.

No.

random photo - update me


ogregoire:
I'm mixed.

Indeed these new damages are added, but they are very specific:

1. The tentacle must hit
2. The creature hit must fail a DC 14 CON saving throw
3. When the creature hit is outside of the water and not moistered, it get these damages every 10 minutes.

This sounds a lot like a script to be made rather than blunt data as is.

So my take on this issue is to have what you wrote, plus having a `condition` (or `special` or whatever) field that is just a copy of the SRD text for the tentacle attack, but starting from the "If the target is a creature, it must succeed on a DC 14 Constitution saving throw or become diseased."

bagelbits:
I remember there are a number monster attacks that are weirdly conditional with some of their damage like this that I skipped over when adding all of the attacks. I know there is a field name we've used previously for this kind of conditional stuff and I'd suggest we'd use the same exact field name if we can, especially since we're going to have to do this in a few places.

ogregoire (approved):

PR #356, opened 2021-05-09
Fix sahuagin psychic

rjb25:
## What does this do?
fix sahuagin json entry

## How was it tested?
It's a simple change. Shouldn't need testing.

Nope

Not requiring documentation

random photo - update me


bagelbits (approved):
LGTM! Good catch!

PR #365, opened 2021-05-21
Add multiclassing models for classes.

Arnoldave:
## What does this do?
This adds multi-classing data to the class definitions. I made this a while ago; thought I might as well throw up a PR, see what you guys think.

## How was it tested?
Same as always; I just build the image in Docker, and then check that the data I added is there.

Not that I know of. Just another case of missing data I spotted.

Not yet; will do it when I make the changes to the API.

random photo - update me


bagelbits:
That looks pretty good. I'm going to let this marinate for a bit and see if anyone else has any feelings on it. The `prerequisite_options` is weird but I get why it's there. Damn Fighters.

Arnoldave:
> That looks pretty good. I'm going to let this marinate for a bit and see if anyone else has any feelings on it. The `prerequisite_options` is weird but I get why it's there. Damn Fighters.

I totally agree with you about prerequisite options. I'm fully open to any suggestions if you come up with any.

bagelbits:
Honestly, I think you've made the right choice. It's just the SRD that's weird.

bagelbits:
Apparently I let it marinate for too long. 😅 

I'll get this merged in.

PR #375, opened 2021-07-04
Tool Proficiency Description Apostrophes

Wyzards:
## What does this do?
The description for the "Tool Proficiency" racial trait used unique Unicode apostrophes which aren't used in any other description, or anywhere else in the database for that matter. These can cause errors when parsing or displaying racial trait descriptions, so they've been replaced with the proper apostrophe character.

## How was it tested?
It wasn't.

I find that highly unlikely.

NA

random photo - update me


bagelbits:
Thanks for finding that!

PR #376, opened 2021-07-05
High Elf Cantrip trait_specific

Wyzards:
## What does this do?
The Elf subrace "High-Elf" has a racial trait called "High Elf Cantrip". This allows them to learn one cantrip from the wizard spell list. Currently, instead of this racial trait being included in high elf's racial_traits attribute, it's included as a choice in high-elf's racial_trait_options attribute. In this choice there are 12 APIReferences that all point back to the same trait. This is redundant and doesn't make use of the trait_specific attribute of the trait like it should.

This change corrects this by doing the following:
- Removes the racial_trait_options attribute from the High Elf subrace
- Adds the trait_specific attribute to the racial trait "high-elf-cantrip"
- Adds the trait "high-elf-cantrip" to the racial_traits attribute of the High Elf subrace

## How was it tested?
It wasn't. I really ought to find out how to run the database locally from my fork...

No

No, I intend to later tonight or tomorrow.

random photo - update me

### DIFF
--- a/src/5e-SRD-Traits.json
+++ b/src/5e-SRD-Traits.json
@@ -283,6 +283,74 @@
       "You know one cantrip of your choice form the wizard spell list. Intelligence is your spellcasting ability for it."
     ],
     "proficiencies": [],
+    "trait_specific": {
+      "wizard_cantrip_options": {
+        "choose": 1,
+        "from": [
+          {
+            "index": "light",
+            "name": "Light",
+            "url": "/api/spells/light"
+          },
+          {
+            "index": "mage-hand",
+            "name": "Mage Hand",
+            "url": "/api/spells/mage-hand"
+          },
+          {
+            "index": "mending",
+            "name": "Mending",
+            "url": "/api/spells/mending"
+          },
+          {
+            "index": "message",
+            "name": "Message",
+            "url": "/api/spells/message"
+          },
+          {
+            "index": "minor-illusion",
+            "name": "Minor Illusion",
+            "url": "/api/spells/minor-illusion"
+          },
+          {
+            "index": "acid-splash",
+            "name": "Acid Splash",
+            "url": "/api/spells/acid-splash"
+          },
+          {
+            "index": "prestidigitation",
+            "name": "Prestidigitation",
+            "url": "/api/spells/prestidigitation"
+          },
+          {
+            "index": "ray-of-frost",
+            "name": "Ray of Frost",
+            "url": "/api/spells/ray-of-frost"
+          },
+          {
+            "index": "shocking-grasp",
+            "name": "Shocking Grasp",
+            "url": "/api/spells/shocking-grasp"
+          },
+          {
+            "index": "true-strike",
+            "name": "True Strike",
+            "url": "/api/spells/true-strike"
+          },
+          {
+            "index": "chill-touch",
+            "name": "Chill Touch",
+            "url": "/api/spells/chill-touch"
+          },
+          {
+            "index": "dancing-lights",
+            "name": "Dancing Lights",
+            "url": "/api/spells/dancing-lights"
+          }
+        ],
+        "type": "spell"
+      }
+    },
     "url": "/api/traits/high-elf-cantrip"
   },
   {
--- a/src/5e-SRD-Traits.json
+++ b/src/5e-SRD-Traits.json
@@ -108,7 +108,7 @@
     "subraces": [],
     "name": "Tool Proficiency",
     "desc": [
-      "You gain proficiency with the artisan’s tools of your choice: smith’s tools, brewer’s supplies, or mason’s tools."
+      "You gain proficiency with the artisan's tools of your choice: smith's tools, brewer's supplies, or mason's tools."
     ],
     "proficiencies": [],
     "proficiency_choices": {
--- a/src/5e-SRD-Races.json
+++ b/src/5e-SRD-Races.json
@@ -405,7 +405,7 @@
         "bonus": 1
       }
     ],
-    "alignment": " Dragonborn tend to extremes, making a conscious choice for one side or the other in the cosmic war between good and evil. Most dragonborn are good, but those who side with evil can be terrible villains.",
+    "alignment": "Dragonborn tend to extremes, making a conscious choice for one side or the other in the cosmic war between good and evil. Most dragonborn are good, but those who side with evil can be terrible villains.",
     "age": "Young dragonborn grow quickly. They walk hours after hatching, attain the size and development of a 10-year-old human child by the age of 3, and reach adulthood by 15. They live to be around 80.",
     "size": "Medium",
     "size_description": "Dragonborn are taller and heavier than humans, standing well over 6 feet tall and averaging almost 250 pounds. Your size is Medium.",
@@ -514,7 +514,7 @@
       }
     ],
     "alignment": "Gnomes are most often good. Those who tend toward law are sages, engineers, researchers, scholars, investigators, or inventors. Those who tend toward chaos are minstrels, tricksters, wanderers, or fanciful jewelers. Gnomes are good-hearted, and even the tricksters among them are more playful than vicious.",
-    "age": " Gnomes mature at the same rate humans do, and most are expected to settle down into an adult life by around age 40. They can live 350 to almost 500 years.",
+    "age": "Gnomes mature at the same rate humans do, and most are expected to settle down into an adult life by around age 40. They can live 350 to almost 500 years.",
     "size": "Small",
     "size_description": "Gnomes are between 3 and 4 feet tall and average about 40 pounds. Your size is Small.",
     "starting_proficiencies": [],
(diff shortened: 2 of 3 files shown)


src/5e-SRD-Traits.json
```diff
@@ -283,6 +283,74 @@
       "You know one cantrip of your choice form the wizard spell list. Intelligence is your spellcasting ability for it."
     ],
     "proficiencies": [],
+    "trait_specific": {
+      "wizard_cantrip_options": {
```
bagelbits, line 287: We might just rename this to `options` for a consistent shape when there is a choice across traits?

Wyzards: Seems reasonable. I was just going off of what @fergcb noted last night over discord regarding trait_specific and feature_specific.

fergcb: I think it's worth distinguishing it in some way, so the consumer knows what to do with the chosen entity. Maybe at least `cantrip_options`.

Or perhaps the `type` of the choice is good enough for that, I don't know.

Wyzards: > I think it's worth distinguishing it in some way, so the consumer knows what to do with the chosen entity. Maybe at least `cantrip_options`.
> 
> Or perhaps the `type` of the choice is good enough for that, I don't know.

Well currently the type is specified as spell. @bagelbits were you referring to the `trait_specific` or the `wizard_cantrip_options` that should've just been renamed to `options`

fergcb: > Seems reasonable. I was just going off of what @fergcb noted last night over discord regarding trait_specific and feature_specific.

`trait_specific` allows us to have a consistent top-level schema for Trait documents, but that doesn't mean the more free-form objects under this attribute can't take advantage of repeating patterns (or patterns that *could* be repeated in theory, e.g. for data from other source books or homebrew).

fergcb: > @bagelbits were you referring to the `trait_specific` or the `wizard_cantrip_options` that should've just been renamed to `options`

One would assume the latter.

Wyzards: > > Seems reasonable. I was just going off of what @fergcb noted last night over discord regarding trait_specific and feature_specific.
> 
> `trait_specific` allows us to have a consistent top-level schema for Trait documents, but that doesn't mean the more free-form objects under this attribute can't take advantage of repeating patterns (or patterns that _could_ be repeated in theory, e.g. for data from other source books or homebrew).

I agree, I just wasn't aware of any such patterns. Should I rename it to options then?

fergcb: I'm inclined to say no to just "options", simply because there's a chance there could be more than one set of "options" (though "choice" feels much more natural here, especially seeing as we call the structure in question a Choice, in the docs) available from one trait, or choices in different traits that do different things (e.g. one that adds a cantrip and one that adds a once-per-rest spell; both would have a Choice `type` of "spell").

If it was just me I'd go for "cantrip_choice", but "cantrip_option" would be fine too.

bagelbits: I could see a case for `spell_options`, but I'm leaning towards `options`?

bagelbits: Yeah. `options` but add in the field `type: spell` into it.

Wyzards: > I'm inclined to say no to just "options", simply because there's a chance there could be more than one set of "options" (though "choice" feels much more natural here, especially seeing as we call the structure in question a Choice, in the docs) available from one trait, or choices in different traits that do different things (e.g. one that adds a cantrip and one that adds a once-per-rest spell; both would have a Choice `type` of "spell").
> 
> If it was just me I'd go for "cantrip_choice", but "cantrip_option" would be fine too.

I see your point here, but consistently when a choice is used, the word options describes it. Never choice. Ex: `racial_trait_options`. `language_options`, `starting_equipment_options`, etc.


I think it should be `options`, `spell_options`, or `cantrip_options`... For consistency, `options` seems best for reusability. `options` being a trait or feature specific attribute that is a choice.

Wyzards: In the event that there are ever multiple trait or feature specific choices, then they can use another name since `options` will have already been taken by the first choice.

fergcb: Sure, the "options" vs. "choice"  thing is mostly a separate issue.

My points about just "options" on its own being too vague still stands I think. I has a chance to clash with other similar, but not semantically identical, attributes.

Wyzards: > Sure, the "options" vs. "choice" thing is mostly a separate issue.
> 
> My points about just "options" on its own being too vague still stands I think. I has a chance to clash with other similar, but not semantically identical, attributes.

I see your point. `spell_options` could work. It's slightly redundant since the type of the choice is already specified as "spell", but I don't think it's too much of an issue. @bagelbits your thoughts?

fergcb: > In the event that there are ever multiple trait or feature specific choices, then they can use another name since `options` will have already been taken by the first choice.

Right, but wouldn't `cantrip_options` and `daily_spell_options` (just for the sake of example be way better than `options` and like, `extra_options` or some other naming convention?

bagelbits: I can see that options being vague. I'm just trying to think about how we'd handle Dragonborn stuff differently. Since there is one choice that affects two other traits.

bagelbits: I'm fine with `spell_options` or `cantrip_options`

Wyzards: > I can see that options being vague. I'm just trying to think about how we'd handle Dragonborn stuff differently. Since there is one choice that affects two other traits.

Well, obviously remove `trait_options` from dragonborn, and add those options as a choice under the `feature_specific` attribute of the dragon type trait. I don't see too much harm in the feature specific choices for the two traits having different names. I also don't have a problem with the choice being named `options` for both and leaving it at that.

Wyzards: The real thing I'm thinking here is that anything under `feature_specific` or `trait_specific` attributes won't be super reusable regardless, since it's literally specific to each trait or feature. I feel that anywhere this information is used it would already need to be individually implemented for that trait. If that's the case then it'd be better for us to try to avoid ambiguousness rather than focus on reusability (since it can't be reused effectively anyways).

bagelbits: Fair enough. And we can also go back if we notice there is reusability.

fergcb: > I can see that options being vague. I'm just trying to think about how we'd handle Dragonborn stuff differently. Since there is one choice that affects two other traits.

This is probably one for another issue/PR, but I think `trait_specific` could handle these sort of "pseudo-traits" like Breath Weapon and Damage Resistance with only one choice, but it would leave more work up to the consumer to handle that specific case.

Maybe having a trait be able to grant other "sub-traits" would simplify things a bit. Similar to the current `trait_options` approach, but contained within the "base" trait's `trait_specific` field instead of the Race document.

Either way, I think the solution is gonna be quite different to that for High Elf cantrips. I wonder if the current mess may have come from trying to treat these two separate issues as two cases of the same issue.

Wyzards: > > I can see that options being vague. I'm just trying to think about how we'd handle Dragonborn stuff differently. Since there is one choice that affects two other traits.
> 
> This is probably one for another issue/PR, but I think `trait_specific` could handle these sort of "pseudo-traits" like Breath Weapon and Damage Resistance with only one choice, but it would leave more work up to the consumer to handle that specific case.
> 
> Maybe having a trait be able to grant other "sub-traits" would simplify things a bit. Similar to the current `trait_options` approach, but contained within the "base" trait's `trait_specific` field instead of the Race document.
> 
> Either way, I think the solution is gonna be quite different to that for High Elf cantrips. I wonder if the current mess may have come from trying to treat these two separate issues as two cases of the same issue.

There is definitely a point to be made here. However, I think the occurences of traits which require you to make a decision are few enough that adding another attribute for subtraits would only further complicate things. I think for the moment it's better to keep it contained within `trait_specific` for both and let it be handled by the consumer.

With that said, @bagelbits is the current naming scheme sufficient or should it be changed to something more general like `spell_options`? The current one definitely has the least ambiguity possible.

fergcb: > I think for the moment it's better to keep it contained within `trait_specific` for both and let it be handled by the consumer.

Absolutely, all of my suggestions above are assuming the data would be under `trait_specific`. The real issue would be somehow defining a Choice structure that has multiple effects; adding a damage resistance *and* an action. Perhaps we ought to discuss this on a separate issue, or in the Discord server.

>  is the current naming scheme sufficient or should it be changed to something more general like `spell_options`?

Another vote for *at least* `cantrip_options`. "spell" is too vague for me. `wizard_cantrip_options` is good, but I can't forsee a circumstance where (for the sake of example) `cleric_cantrip_options` would be treated differently by the consumer. Then again, just because I can't foresee it doesn't mean it's impossible.

Like @bagelbits said before, if we go for higher specificity now, we can always backtrack if we ever discover a more generic repeating pattern.

Wyzards: > Absolutely, all of my suggestions above are assuming the data would be under `trait_specific`. The real issue would be somehow defining a Choice structure that has multiple effects; adding a damage resistance _and_ an action. Perhaps we ought to discuss this on a separate issue, or in the Discord server

I see, I misunderstood you there. In that case I agree, it's something we should discuss elsewhere. For now, this works.
> Another vote for _at least_ `cantrip_options`. "spell" is too vague for me. `wizard_cantrip_options` is good, but I can't forsee a circumstance where (for the sake of example) `cleric_cantrip_options` would be treated differently by the consumer. Then again, just because I can't foresee it doesn't mean it's impossible.
> 
> Like @bagelbits said before, if we go for higher specificity now, we can always backtrack if we ever discover a more generic repeating pattern.

Alright with that in mind I think `wizard_cantrip_options` is fine for now and we'll change it later if we have to. With that said, I don't think there's any other issues preventing a merge?

bagelbits (approved):
Lgtm!

PR #377, opened 2021-07-05
Replace the non-ASCII characters with ASCII equivalent.

ogregoire:
## What does this do?
It replaces all the non-ASCII characters with their ASCII counterpart, except for `×` because it's the only non-ASCII character without any strict equivalent (I even replaced a few `x` with `×`. So now, the only non-ASCII character in the data is `×`, which is the *multiply* character, useful in mathematical formulae such as `1d10 × 100 feet`.

## How was it tested?
Tested with the regex  `[^ -~\n×]` in the JSON files.

No.

The change doesn't require any modification in the docs.

random photo - update me


bagelbits (approved):
Thank for catching these!

PR #378, opened 2021-07-07
Feature trait choice

Wyzards:
## What does this do?
- Replaces the attribute `group` of features with `parent.` It's type is APIReference, and it points to the parent feature.
- For any features with the `choice` attribute, it has been moved within the newly created attribute `feature_specific`. It has also been renamed according to what the choice entails. Also, any choices that used features when they could have used other data structures (ex: proficiency for expertise instead of a feature), have had their excess features removed and their choice data type changed.
- Applies the previous changes to the `draconic-ancestry` racial trait, and removes the `trait_options` attribute from the dragonborn rave.

## How was it tested?
Wasn't yet. Intend to run tests on locally run db.

## Is there a Github issue this is resolving?
No.

## Did you update the docs in the API? Please link an associated PR if applicable.
Not yet, I intend to.

## Here's a fun image for your troubles
random photo - update me

PR #380, opened 2021-07-08
Feature & Trait Choices

Wyzards:
## What does this do?
- Replaces Feature's `group` attribute with the `parent` attribute. Its type is APIReference, and it points to the parent feature.
- Moves Feature's `choice` attribute into Feature's `feature_specific` attribute. Also renames it (either to `subfeature_options`, `expertise_options`, `spell_options` or `trait_options` for trait equivalents.
- Changes choice type of `rogue-choose-expertise` and `bard-choose-expertise` to proficiency.
- Removes unnecessary bard and rogue expertise features.
- Adds a choice structure to the trait `draconic_ancestry`'s feature specific. Also includes `draconic_ancestry`'s damage type and breath weapon action under `trait_specific`
- Removes "Choose: " from the name of Features which involve making a choice.
- removes "choose-" and "-choose" from traits / features whos index includes it.

These changes were discussed in the API Discord between myself and @fergcb with some input from @bagelbits and @ogregoire in a different draft PR.

Discord discussion link: https://discord.com/channels/656547667601653787/656547667601653790/861810226541625395
Old draft PR link: https://github.com/5e-bits/5e-database/pull/378

## How was it tested?
Database was run locally on machine. Was built from remote git branch and ran tests according to api documentation (unit test, integration and controllers).

No.

Yes, https://github.com/5e-bits/5e-srd-api/pull/226

random photo - update me


ogregoire:
This is good, and it was well needed. Your correction regarding the previous PR is accurate and makes the naming more consistent regarding previous changes (removing "Choose:" rather than adding it, even though we should go further on that topic). I no longer want this PR to be split into several different as it does two things but does them well.

Since the PR is in draft mode, I cannot approve it in the GitHub sense of it, but I'm approving it here in written.

Given the large amount of commits, I believe that this PR should be squash-merged. What's your opinion on this?

src/5e-SRD-Features.json
```diff
@@ -1117,107 +759,207 @@
       "name": "Bard",
       "url": "/api/classes/bard"
     },
-    "name": "Choose: Expertise 2",
+    "name": "Expertise 2",
```
fergcb, line 762: This could be something for a different PR, and I know it's not something you've introduced in this PR, but I'm wondering whether names like "Expertise (10th Level)" would be more appropriate than "Expertise 2"? We could also cut down the description to only include the effects that happen at this level, perhaps?

src/5e-SRD-Features.json
```diff
@@ -1117,107 +759,207 @@
...
+            "name": "Skill: Sleight of Hand",
+            "url": "/api/proficiencies/skill-sleight-of-hand"
+          },
+          {
+            "index": "skill-stealth",
+            "name": "Skill: Stealth",
+            "url": "/api/proficiencies/skill-stealth"
+          },
+          {
+            "feature_specific": {
```
fergcb: This seems like it shouldn't be here?

src/5e-SRD-Features.json
```diff
@@ -2025,7 +1769,11 @@
     "desc": [
       "The Circle of the Land is made up of mystics and sages who safeguard ancient knowledge and rites through a vast oral tradition. These druids meet within sacred circles of trees or standing stones to whisper primal secrets in Druidic. The circle's wisest members preside as the chief priests of communities that hold to the Old Faith and serve as advisors to the rulers of those folk. As a member of this circle, your magic is influenced by the land where you were initiated into the circle's mysterious rites."
     ],
-    "group": "Circle of the Land",
+    "parent": {
+      "index": "choose-circle-of-the-land",
```
fergcb: Ought we to update the index to match the name, i.e. removing `choose-`?

fergcb:
> Given the large amount of commits, I believe that this PR should be squash-merged. What's your opinion on this?

All multi-commit PRs in this repo are squashed by default, anyway, are they not?

src/5e-SRD-Features.json
ogregoire: I guess that the change is small for now, and we should keep as is. We can discuss this in greater lengths, see #382 for that. Adding this request would increase the inconsistency or, if done entirely, add more complexity to this PR.

src/5e-SRD-Features.json
ogregoire: Indeed, is this a bad copy/paste?

src/5e-SRD-Features.json
ogregoire: This is also mentioned in #382 .

bagelbits:
> All multi-commit PRs in this repo are squashed by default, anyway, are they not?

@fergcb @ogregoire That is correct. All PRs are `Squash and Merge`. It makes it easy to rollback/revert for both repos if something goes wrong.

src/5e-SRD-Features.json
bagelbits: Yeah. I think keeping it as is for now makes sense and we can make a separate PR to address the discussion in #382

src/5e-SRD-Features.json
Wyzards: > Ought we to update the index to match the name, i.e. removing `choose-`?

I agree with removing choose from the indexes. Should I do that in this pr or another one?

fergcb: If the names are changing in this one, I think the indexes should, too

Wyzards: Should that include everything in https://github.com/5e-bits/5e-database/discussions/382 then as well? Or just removing choose from indexes.

fergcb: I reckon leave the rest for now. If we've decided to do the names in this PR, it feels like the indices should be done now too, just for consistency.

bagelbits (approved):
Double approve since @ogregoire approved it too.

PR #384, opened 2021-07-12
SRD Accuracy and Redundancy For Features

Wyzards:
## What does this do?
This pull request makes several changes with the goal of eliminating redundancy within various features and traits, while also attempting to make them more accurate to the D&D 5e System Reference Document. The changes are as follows:

1. Some features grant additional benefits at level levels from when they're initially gained. For example, Bard's Expertise feature is granted at 3rd level. However, it grants an additional benefit (more expertise) at 10th level. While in the SRD this is noted as one individual feature, we include one for each benefit within the API so that we can better detail class leveling perks. The features are named `bard-expertise-1` and `bard-expertise-2` accordingly. Many features that have these naming schemes include the # in their name. This has been removed to be more accurate to the SRD.
2. Some other features that grant additional benefits at later levels use an "additional" naming scheme. An example is the current name for Eldritch Invocations. The first Eldritch Invocations feature has an index of `eldritch-invocation`, with the second one being `additional-eldritch-invocation-1`. Any indexes that "additional" without it being in the feature name in the SRD have had it removed, and have had their number incremented to match the original feature. (New names would be `eldritch-invocations-1`, `eldritch-invocations-2`, and so on...)
3. Most features with additional benefits duplicate the description of their original feature. This is redundant since the feature essentially points to the original (parent) feature. As such, the description has been removed from these additional features. The `parent` attribute has also been given to these features if not already possessed, which is an APIReference that points to the original feature. (Ex: `bard-expertise-2` has a `parent` attribute that is an APIReference to `bard-expertise-1`).
4. The Ability Score Improvement feature is shared by many classes. Every class except for rogue and fighter uses the same Ability Score Improvement feature in the SRD. Despite this, there is a unique feature for each class in the API, with the exact same information. This repetition of information is redundant and as such the 10 sharing classes have had their Ability Score Improvement features condensed into one (actually 5 for each level they are obtained at). They're named `ability-score-improvement-1`, `ability-score-improvement-2`, etc. The features for Fighter and Rogue are unchanged.
5. The `draconic-ancestry` trait of Dragonborns can make use of the following changes, so they have been implemented for it as well.

Some people may disagree with the 3rd changes mentioned. As such, all of the commits for it have been made separately so the other commits can be cherry picked if need be.

**Important Structural Changes:**
- There now exists some features and traits with no `desc` attribute. This is unprecedented as of now.
- There is now a `classes` attribute within Features (currently is only used by Ability Score Improvements, but there may be other applicable uses), for when a feature is used by multiple classes. It is a list of Class APIReferences.

## How was it tested?
As this is a draft PR, it has not been tested yet. Tomorrow I will build the API locally and run the typical tests on it.

There is not an issue, but there is a discussion that discussed changes 1 and 2. Link: https://github.com/5e-bits/5e-database/discussions/382

I will update the documentation for this before I make this PR ready for review.

random photo - update me

### DIFF
--- a/src/5e-SRD-Traits.json
+++ b/src/5e-SRD-Traits.json
@@ -524,9 +524,6 @@
     ],
     "subraces": [],
     "name": "Draconic Ancestry (Black)",
-    "desc": [
-      "You have draconic ancestry. Choose one type of dragon from the Draconic Ancestry table. Your breath weapon and damage resistance are determined by the dragon type, as shown in the table."
-    ],
     "parent": {
       "index": "draconic-ancestry",
       "name": "Draconic Ancestry",
@@ -584,9 +581,6 @@
     ],
     "subraces": [],
     "name": "Draconic Ancestry (Blue)",
-    "desc": [
-      "You have draconic ancestry. Choose one type of dragon from the Draconic Ancestry table. Your breath weapon and damage resistance are determined by the dragon type, as shown in the table."
-    ],
     "parent": {
       "index": "draconic-ancestry",
       "name": "Draconic Ancestry",
@@ -644,9 +638,6 @@
     ],
     "subraces": [],
     "name": "Draconic Ancestry (Brass)",
-    "desc": [
-      "You have draconic ancestry. Choose one type of dragon from the Draconic Ancestry table. Your breath weapon and damage resistance are determined by the dragon type, as shown in the table."
-    ],
     "proficiencies": [],
     "trait_specific": {
       "damage_type": {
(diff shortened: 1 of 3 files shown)


src/5e-SRD-Traits.json
```diff
@@ -524,9 +524,6 @@
     ],
     "subraces": [],
     "name": "Draconic Ancestry (Black)",
-    "desc": [
```
fergcb, line 527: Given that all other traits have descriptions, it would probably make more sense to make this an empty array rather than removing it completely, so the schema is consistent.

Wyzards: I have no issues with this. I'll implement the change tomorrow if I have the time.

ogregoire:
Again this PR does too many things: I like 1 and 2, but I dislike 3 and I'm more or less in favor of 4 but I guess it should be refined. The 5 is a trait and should be done separately.

I see that you're enthusiastic about this, @Wyzards, and I thank you for that. But I really believe that we should take the time to discuss. Do not hesitate to open discussions about what you'd like to see changed. I'm not opposed to change, I just want to make sure that if we change something, it's for the better. I see issues with 3, 4 and 5, and I would like to address those issues before merging. Unfortunately, given that everything is in this PR, it'll be hard to discuss them or to make sure that we covered all cases for each of those ideas.

PR #385, opened 2021-07-14
Name/Index Discussion changes

Wyzards:
## What does this do?
There was a brief discussion about names and indexes of certain features not being accurate to the SRD and it seemed to be agreed that the proposed changes were good. I have implemented them. Here is the relevant discussion link: https://github.com/5e-bits/5e-database/discussions/382

It adds the following:

1. Some features grant additional benefits at level levels from when they're initially gained. For example, Bard's Expertise feature is granted at 3rd level. However, it grants an additional benefit (more expertise) at 10th level. While in the SRD this is noted as one individual feature, we include one for each benefit within the API so that we can better detail class leveling perks. The features are named bard-expertise-1 and bard-expertise-2 accordingly. Many features that have these naming schemes include the # in their name. This has been removed to be more accurate to the SRD.
2. Some other features that grant additional benefits at later levels use an "additional" naming scheme. An example is the current name for Eldritch Invocations. The first Eldritch Invocations feature has an index of eldritch-invocation, with the second one being additional-eldritch-invocation-1. Any indexes that "additional" without it being in the feature name in the SRD have had it removed, and have had their number incremented to match the original feature. (New names would be eldritch-invocations-1, eldritch-invocations-2, and so on...)

Previous draft PR (some commits were cherry picked into this one): https://github.com/5e-bits/5e-database/pull/384

## How was it tested?
It wasn't.

No, but the relevant discussion is linked.

Not needed.

random photo - update me

### DIFF
--- a/src/5e-SRD-Levels.json
+++ b/src/5e-SRD-Levels.json
@@ -1334,7 +1334,7 @@
       },
       {
         "index": "domain-spells-1",
-        "name": "Domain Spells 1",
+        "name": "Domain Spells",
         "url": "/api/features/domain-spells-1"
       }
     ],
@@ -1413,7 +1413,7 @@
     "features": [
       {
         "index": "domain-spells-2",
-        "name": "Domain Spells 2",
+        "name": "Domain Spells",
         "url": "/api/features/domain-spells-2"
       }
     ],
@@ -1487,7 +1487,7 @@
     "features": [
       {
         "index": "domain-spells-3",
-        "name": "Domain Spells 3",
+        "name": "Domain Spells",
         "url": "/api/features/domain-spells-3"
       },
       {
@@ -1566,7 +1566,7 @@
     "features": [
       {
         "index": "domain-spells-4",
-        "name": "Domain Spells 4",
+        "name": "Domain Spells",
         "url": "/api/features/domain-spells-4"
       }
     ],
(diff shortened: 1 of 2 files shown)


Wyzards:
@ogregoire hopefully this is more manageable. I'll get to the other PRs when I can. Am out of state ATM.

ogregoire (approved):

src/5e-SRD-Levels.json
```diff
@@ -6887,9 +6887,9 @@
     "prof_bonus": 3,
     "feature_choices": [
       {
-        "index": "additional-eldritch-invocation-1",
-        "name": "Additional Eldritch Invocation",
-        "url": "/api/features/additional-eldritch-invocation-1"
+        "index": "eldritch-invocations-2",
```
bagelbits, line 6890: Should these be off by one compared to the original indices? I'm sorry I'm on my phone or I would be able to check properly.

bagelbits (commented):
A small comment but otherwise looks good!

src/5e-SRD-Levels.json
Wyzards: > Should these be off by one compared to the original indices? I'm sorry I'm on my phone or I would be able to check properly.

They should be. The first eldritch invocations feature was just named `eldritch-invocations`. It is now `eldritch-invocations-1` and all subsequent additional eldritch invocations follow the same naming scheme, hence the incriment.

PR #386, opened 2021-07-16
Ability Score Improvement Redundancy

Wyzards:
## What does this do?
The Ability Score Improvement feature is shared by many classes. Every class except for rogue and fighter uses the same Ability Score Improvement feature in the SRD. Despite this, there is a unique feature for each class in the API, with the exact same information. This repetition of information is redundant and as such the 10 sharing classes have had their Ability Score Improvement features condensed into one (actually 5 for each level they are obtained at). They're named `ability-score-improvement-1`, `ability-score-improvement-2`, etc. The features for Fighter and Rogue are unchanged.

Currently features have a `class` attribute which indicates which class has the feature. Currently this supports only one class reference. As such, I've given the new ability score improvement features an attribute called `classes` which is just an array of class APIReferences.

## How was it tested?
It wasn't.

No

N/A.

random photo - update me

### DIFF
--- a/src/5e-SRD-Features.json
+++ b/src/5e-SRD-Features.json
@@ -101,21 +101,6 @@
     "url": "/api/features/frenzy"
   },
   {
-    "index": "barbarian-ability-score-improvement-1",
-    "class": {
-      "index": "barbarian",
-      "name": "Barbarian",
-      "url": "/api/classes/barbarian"
-    },
-    "name": "Barbarian: Ability Score Improvement 1",
-    "level": 4,
-    "prerequisites": [],
-    "desc": [
-      "When you reach 4th level, and again at 8th, 12th, 16th, and 19th level, you can increase one ability score of your choice by 2, or you can increase two ability scores of your choice by 1. As normal, you can't increase an ability score above 20 using this feature."
-    ],
-    "url": "/api/features/barbarian-ability-score-improvement-1"
-  },
-  {
     "index": "barbarian-extra-attack",
     "class": {
       "index": "barbarian",
@@ -182,21 +167,6 @@
     "url": "/api/features/feral-instinct"
   },
   {
-    "index": "barbarian-ability-score-improvement-2",
-    "class": {
-      "index": "barbarian",
-      "name": "Barbarian",
-      "url": "/api/classes/barbarian"
-    },
-    "name": "Barbarian: Ability Score Improvement 2",
-    "level": 8,
-    "prerequisites": [],
-    "desc": [
-      "When you reach 4th level, and again at 8th, 12th, 16th, and 19th level, you can increase one ability score of your choice by 2, or you can increase two ability scores of your choice by 1. As normal, you can't increase an ability score above 20 using this feature."
-    ],
-    "url": "/api/features/barbarian-ability-score-improvement-2"
-  },
-  {
     "index": "brutal-critical-1-die",
     "class": {
       "index": "barbarian",
@@ -249,21 +219,6 @@
     "url": "/api/features/relentless-rage"
   },
   {
-    "index": "barbarian-ability-score-improvement-3",
-    "class": {
-      "index": "barbarian",
-      "name": "Barbarian",
-      "url": "/api/classes/barbarian"
-    },
-    "name": "Barbarian: Ability Score Improvement 3",
-    "level": 12,
-    "prerequisites": [],
-    "desc": [
-      "When you reach 4th level, and again at 8th, 12th, 16th, and 19th level, you can increase one ability score of your choice by 2, or you can increase two ability scores of your choice by 1. As normal, you can't increase an ability score above 20 using this feature."
-    ],
-    "url": "/api/features/barbarian-ability-score-improvement-3"
-  },
-  {
     "index": "brutal-critical-2-dice",
     "class": {
       "index": "barbarian",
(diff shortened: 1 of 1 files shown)


Wyzards:
@ogregoire you mentioned in https://github.com/5e-bits/5e-database/pull/384 that this change could be refined. What would you suggest?

ogregoire:
First, there's a bug: in the current presentation, the barbarian has no way of having any ASI because it's forgotten in "ability-score-improvement-X".

## Why this change?

But more broadly, I question the need for such requirement of moving the ASI in only one place. I don't see the issue with the current implementation. Why is redundancy bad? Are we creating a [6NF database](https://en.wikipedia.org/wiki/Database_normalization)? (No, we aren't). And redundancy is present everywhere. We've always preferred to have things being laid out explicitly rather than implicitly and this change you suggest, introduce a change in that philosophy. While it's not written in stone, it's the idea when you look at the whole database. So if there is such a change to reduce clutter, I'm open to do it an senseful manner.

## This change makes extensions hard to implement.

The advantage of laying out things explicitly is that it's entirely possible to extend the database much more easily, it makes it more open to customization by others and less prone to side effects.

For instance, I create a "mod" that says "Druids can take an extra ASI" at level 9, and want to mention it in the druid text. As a modder, it makes sense that I'm gonna have to change the druid ASI instances, and add mine. But now, with the proposal you make, I now have to create myself all the druid ASI **and** also modify the classes of the main ASI to **remove** druid from it. This is counter-intuitive.

Another example, if anyone wants to use the D&D Beyond route, and change the text for each ASI to the following, they now can't without remove the new ASI implementation and redo the old one by themselves, where before I imagine people would simply overwrite the description text.

> Level 4: "When you reach 4th level, and again at 6th, 8th, 12th, 14th, 16th, and 19th level"

> Level 8: "When you reach 8th level, and again at 12th, 14th, 16th, and 19th level"

One might question the need for extension, but if you look at why the SRD simply exists ([check WotC's own FAQ](https://dnd.wizards.com/articles/features/systems-reference-document-srd)), it's to be the base that is extensible. The very existence of the SRD is to make D&D extensible by everyone and not only by WotC. Check each new sourcebook to see that D&D wants extension to their system every time. If one cannot easily extend, I don't see the point of this project. So to me, this makes extensions way harder to implement in regards to ASI changes for what gain? Less clutter? Not worth it, IMO.

## Technically, it should be improved.

Another issue is the change from `class` to `classes` but only for some features. This is unacceptable. We have two fields that mean exactly the same thing except one is a list and one is a single item. This duality between `class` and `classes` increases the complexity where we never did before. If this change must go live, all features should have `classes` rather than `class` and use the list/array structure.

## Now what?

So I'm not against reducing clutter, but I believe it should be done in a smarter way than what is presented here. I haven't found anything that works better than what exists now, so if you want to rethink this idea to make it less explicit while keeping its openness and philosophy, I'm all ears.

But on a technical level, the barbarian must be added, and the `class` attribute should be replaced with `classes` (and a list) everywhere.

Wyzards:
> First, there's a bug: in the current presentation, the barbarian has no way of having any ASI because it's forgotten in "ability-score-improvement-X".

This escaped my notice, I'll fix it.

> But more broadly, I question the need for such requirement of moving the ASI in only one place. I don't see the issue with the current implementation. Why is redundancy bad? Are we creating a [6NF database]

In my last semester of College I had to do a database class and it involved a focus on normalization and cutting out redundancy. I suppose it's more of a habit than anything at this point and I didn't question whether or not it should be included.

> And redundancy is present everywhere. We've always preferred to have things being laid out explicitly rather than implicitly 

Can I have an example of where this philosophy is in play, if any are known?

> ## This change makes extensions hard to implement.
> The advantage of laying out things explicitly is that it's entirely possible to extend the database much more easily, it makes it more open to customization by others and less prone to side effects.
> 
> For instance, I create a "mod" that says "Druids can take an extra ASI" at level 9, and want to mention it in the druid text. As a modder, it makes sense that I'm gonna have to change the druid ASI instances, and add mine. But now, with the proposal you make, I now have to create myself all the druid ASI **and** also modify the classes of the main ASI to **remove** druid from it. This is counter-intuitive.

You would still have to create a new ASI feature for the specified level with the current system, but I suppose creating 1 feature is less than creating 5. Valid point, however I don't think it's particularly relevant since there's not a situation where we would ever have to create such a modification? Nontheless, valid.

> Another example, if anyone wants to use the D&D Beyond route, and change the text for each ASI to the following, they now can't without remove the new ASI implementation and redo the old one by themselves, where before I imagine people would simply overwrite the description text.
> 
> > Level 4: "When you reach 4th level, and again at 6th, 8th, 12th, 14th, 16th, and 19th level"
> 
> > Level 8: "When you reach 8th level, and again at 12th, 14th, 16th, and 19th level"

I'm not quite sure what you mean by this.

> ## Technically, it should be improved.
> Another issue is the change from `class` to `classes` but only for some features. This is unacceptable. We have two fields that mean exactly the same thing except one is a list and one is a single item. This duality between `class` and `classes` increases the complexity where we never did before. If this change must go live, all features should have `classes` rather than `class` and use the list/array structure.

I agree with this, and is something I'd be willing to change.

> ## Now what?
> So I'm not against reducing clutter, but I believe it should be done in a smarter way than what is presented here. I haven't found anything that works better than what exists now, so if you want to rethink this idea to make it less explicit while keeping its openness and philosophy, I'm all ears.
> 
> But on a technical level, the barbarian must be added, and the `class` attribute should be replaced with `classes` (and a list) everywhere.

I agree. What you say about removing clutter not being super necessary, especially for a potential decrease in extendibility, rings true. If you can show some examples of us not prioritizing redundancy then I'll likely drop this change. Otherwise I'll start by making the mentioned changes.

bagelbits:
> In my last semester of College I had to do a database class and it involved a focus on normalization and cutting out redundancy. I suppose it's more of a habit than anything at this point and I didn't question whether or not it should be included.

Normalization is the right thing to do with SQL databases, but NoSQL is a different beast. Consistent schema is more the name of the game here.

Also, this change doesn't exactly work because Rogue and Fighter are slightly different than the other classes. Rogue gets an extra ASI at 10th level and Fighter gets an extra ASI at 6th level.

I appreciate the optimization that you're attempting here but I don't think it quite works. You will still have to make 3 different sets of ASI entries: a general one and specific ones for Rogue and Fighter.

I think we should close out this PR for now because it doesn't feel like the right fit. Thank you for the time and thought you put into this though.

PR #387, opened 2021-07-16
Missing levels ASI names

Wyzards:
## What does this do?
The Ability Score Improvement feature was incorrectly named in the levels file for each class. It was named as "Ability Score Improvement 1", "Ability Score Improvement 2", etc. and has now had the number removed for all instances. This was meant to be included in https://github.com/5e-bits/5e-database/pull/385 but it seems the commit didn't push.

## How was it tested?
It wasn't.

There is not.

Not applicable.

random photo - update me


bagelbits (approved):
LGTM!

PR #389, opened 2021-07-18
Added Quantity To Bard Lute

Wyzards:
## What does this do?
Adds a missing "quantity" attribute to the Lute in Bard's starting-equipment-options.

## How was it tested?
It wasn't.

No.

NA

random photo - update me


bagelbits (approved):
LGTM. I think we can drop the starting equipment table though.

PR #398, opened 2021-07-29
add Melee Weapons and Ranged Weapons categories

MikkelPaulson:
## What does this do?
We already synthesize the "(Simple|Martial) (Melee|Ranged) Weapons" categories into "Simple Weapons" and "Martial Weapons", so it doesn't seem unreasonable to also have mashed-up categories for melee and ranged.

**Outstanding question:** Should melee weapons with the thrown property be included in both categories? I'm conflicted about this one, but as the PR stands now the melee and ranged categories are exclusively combinations of their simple and martial subsets.

## How was it tested?
* Manually tested in Docker container.
* Manually ran Jest.

No.

N/A

### DIFF
--- a/src/5e-SRD-Equipment-Categories.json
+++ b/src/5e-SRD-Equipment-Categories.json
@@ -1791,6 +1791,205 @@
     "url": "/api/equipment-categories/martial-weapons"
   },
   {
+    "index": "melee-weapons",
+    "name": "Melee Weapons",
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
LGTM.

> Should melee weapons with the thrown property be included in both categories?

No. I think not if we're mirroring how things are stored in the SRD.

Issue #145, opened 2020-01-31
CI Test for unformity


benjaminapetersen:
Thinking probably a mocha/chai script (or just plain ol' javascript really, no need to go crazy):

- https://mochajs.org/#getting-started
- https://www.chaijs.com/

First could be something simple like

```javascript 
describe("monsters", function() => {
    describe('JSON consistency', function() {
        it('should follow a consistent structure', function() {
           monsters.forEach((monster) => {
              expect(monster.name).to.be.a('string');
              expect(monster.size).to.be.a('string');
              expect(monster.armor_class).to.be.a('number');
              // etc etc.
              // more valuable for complex properties.
           });
      }); 
  });
});
```

benjaminapetersen:
@bagelbits @ogregoire been off a bit but thinking about this one, since there are a lot of changes lately.  changes are great! This would ensure they are uniform.   Down side is the test has to be updated with the changes (but so long as they stay fairly simple, this is prob an upside as well).

bagelbits:
I mean this would lead towards more standardized Schemas, which would also make life easier to document them in the API as well. (I could definitely use help with documenting stuff but also it's kind of on me to do it).

brittonhayes:
Just wanted to hop on this thread to say this would make it much easier for folks like me to maintain our API clients for the DnD project. Some sort of consistent schema checking is definitely a bigtime win in my book.

benjaminapetersen:
I offered this other comment which likely applies here:

https://github.com/5e-bits/5e-database/issues/1006#issuecomment-4076182318

But depends on the community willingness to utilize the suggested technology.

bagelbits:
I believe we have this covered now with a ZOD schema.

PR #399, opened 2021-07-31
Test to enforce consistency of references

MikkelPaulson:
## What does this do?
I've been paranoid about making a typo in my `"index"`/`"name"`/`"url"` while adding references, so I added a test to verify that a) all referenced URLs are present as top-level dicts somewhere in the project, and b) the names and indexes, if present, match those of the referenced entities.

This led me down a rabbit hole of cleaning up all of the failures identified by the test. Most were capitalization related, where I preferred Title Case in all cases, but it did turn up a few interesting data errors too. Tieflings don't speak Abyssal, Drow and Dryders don't cast Entangle, and Rakshasas *definitely* don't cast Mass Suggestion.

I'm not a JS programmer, so the test is pretty hackish, but it works.

## How was it tested?
I ran it.

No.

N/A

external-content duckduckgo com

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -13327,7 +13327,7 @@
           "components_required": [],
           "spells": [
             {
-              "name": "Entangle",
+              "name": "Dancing Lights",
               "level": 1,
               "url": "/api/spells/dancing-lights",
               "usage": {
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -29589,7 +29589,7 @@
             {
               "name": "Suggestion",
               "level": 2,
-              "url": "/api/spells/mass-suggestion",
+              "url": "/api/spells/suggestion",
               "usage": {
                 "type": "per day",
                 "times": 3
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -13571,7 +13571,7 @@
             {
               "name": "Darkness",
               "level": 2,
-              "url": "/api/spells/entangle",
+              "url": "/api/spells/darkness",
               "usage": {
                 "type": "per day",
                 "times": 1
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -13882,7 +13882,7 @@
               }
             },
             {
-              "name": "Pass without Trace",
+              "name": "Pass Without Trace",
               "level": 2,
               "url": "/api/spells/pass-without-trace",
               "usage": {
@@ -35748,7 +35748,7 @@
               }
             },
             {
-              "name": "Pass without Trace",
+              "name": "Pass Without Trace",
               "level": 2,
               "url": "/api/spells/pass-without-trace",
               "usage": {
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -37014,7 +37014,7 @@
   },
   {
     "index": "werebear-bear",
-    "name": "Werebear, Bear form",
+    "name": "Werebear, Bear Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37128,7 +37128,7 @@
   },
   {
     "index": "werebear-human",
-    "name": "Werebear, Human form",
+    "name": "Werebear, Human Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37226,7 +37226,7 @@
   },
   {
     "index": "werebear-hybrid",
-    "name": "Werebear, Hybrid form",
+    "name": "Werebear, Hybrid Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37374,7 +37374,7 @@
   },
   {
     "index": "wereboar-boar",
-    "name": "Wereboar, Boar form",
+    "name": "Wereboar, Boar Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37465,7 +37465,7 @@
   },
   {
     "index": "wereboar-human",
-    "name": "Wereboar, Human form",
+    "name": "Wereboar, Human Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37568,7 +37568,7 @@
   },
   {
     "index": "wereboar-hybrid",
-    "name": "Wereboar, Hybrid form",
+    "name": "Wereboar, Hybrid Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37702,7 +37702,7 @@
   },
   {
     "index": "wererat-human",
-    "name": "Wererat, Human form",
+    "name": "Wererat, Human Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -37842,7 +37842,7 @@
   },
   {
     "index": "wererat-hybrid",
-    "name": "Wererat, Hybrid form",
+    "name": "Wererat, Hybrid Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38021,7 +38021,7 @@
   },
   {
     "index": "wererat-rat",
-    "name": "Wererat, Rat form",
+    "name": "Wererat, Rat Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38112,7 +38112,7 @@
   },
   {
     "index": "weretiger-human",
-    "name": "Weretiger, Human form",
+    "name": "Weretiger, Human Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38241,7 +38241,7 @@
   },
   {
     "index": "weretiger-hybrid",
-    "name": "Weretiger, Hybrid form",
+    "name": "Weretiger, Hybrid Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38411,7 +38411,7 @@
   },
   {
     "index": "weretiger-tiger",
-    "name": "Weretiger, Tiger form",
+    "name": "Weretiger, Tiger Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38521,7 +38521,7 @@
   },
   {
     "index": "werewolf-human",
-    "name": "Werewolf, Human form",
+    "name": "Werewolf, Human Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38633,7 +38633,7 @@
   },
   {
     "index": "werewolf-hybrid",
-    "name": "Werewolf, Hybrid form",
+    "name": "Werewolf, Hybrid Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
@@ -38751,7 +38751,7 @@
   },
   {
     "index": "werewolf-wolf",
-    "name": "Werewolf, Wolf form",
+    "name": "Werewolf, Wolf Form",
     "size": "Medium",
     "type": "humanoid",
     "subtype": "human",
--- a/src/tests/tables.test.js
+++ b/src/tests/tables.test.js
@@ -21,7 +21,7 @@ describe("duplicate indices", () => {
   });
 });
 
-describe.skip("broken links", () => {
+describe("broken links", () => {
   it("should not contain broken links", () => {
     let errors = [];
     let files = glob.sync("src/*.json");
--- a/src/tests/tables.test.js
+++ b/src/tests/tables.test.js
@@ -39,20 +39,22 @@ describe("api references", () => {
       resources[entry.url] = { index: entry.index, name: entry.name };
     });
 
-    forEachFileRecursive((filename, entry) => {
-      if (!entry.hasOwnProperty("url")) return;
+    forEachFile((filename, topLevelEntry) => {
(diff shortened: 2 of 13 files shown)


bagelbits:
I'll review this later but holy crap thank you! I think there's an open issue for exactly this.

src/5e-SRD-Monsters.json
```diff
@@ -29589,7 +29589,7 @@
             {
               "name": "Suggestion",
               "level": 2,
-              "url": "/api/spells/mass-suggestion",
+              "url": "/api/spells/suggestion",
```
bagelbits, line 29592: It's actually neither of these and is `shillelagh`

src/5e-SRD-Monsters.json
```diff
@@ -35748,7 +35748,7 @@
               }
             },
             {
-              "name": "Pass without Trace",
+              "name": "Pass Without Trace",
```
bagelbits, line 35751: Looks like this is a repeat? Dryad might just have their spells entirely wrong.

src/5e-SRD-Monsters.json
MikkelPaulson: Just checked the SRD 5.1 for Rakshasa, it lists the following:

> * At will: detect thoughts, disguise self, mage hand, minor illusion
> * 3/day each: charm person, detect magic, invisibility, major image, **suggestion**
> * 1/day each: dominate person, fly, plane shift, true seeing

The rest of the spells look correct.

src/5e-SRD-Monsters.json
MikkelPaulson: This block is for Unicorn. Verified against the SRD, looks correct to me.

MikkelPaulson (commented):
It definitely brushes up against #145, but I'm not enforcing any particular schema here, just verifying that fields are consistent if present.

Two things that I'm explicitly not checking, to keep the scope of cleanup changes small(er), are:

1. Are the `"index"` and `"name"` fields present? (`5e-SRD-Monsters.json` in particularly is sometimes missing `"index"` and sometimes uses `"id"` instead.)
2. Does the `"index"` field match the trailing portion of the `"url"`? (The only violations of this one seem deliberate, eg. `"/api/subclasses/draconic/levels/14"` is indexed as `"draconic-14"`.)

src/5e-SRD-Monsters.json
bagelbits: For some reason I thought it was Dryad. My bad.

src/tests/tables.test.js
```diff
@@ -20,3 +20,60 @@ describe("duplicate indices", () => {
...
+        } else if (typeof object[property] === "object") {
+          walk_recursive(resources, file, object[property], errors);
+        }
+      }
+    };
+
+    for (const file of files) {
+      const fileText = fs.readFileSync(file, "utf8");
+      const fileJSON = JSON.parse(fileText);
+      fileJSON.forEach((entry) => {
```
bagelbits: You might be able to extract this into a helper method? Something like this:
```js
const forEachFile(callback) {
  for (const file of files) {
      const fileText = fs.readFileSync(file, "utf8");
      const fileJSON = JSON.parse(fileText);
      fileJSON.forEach(callback);
  }
}
```

This would allow you to just say:
```js
forEachFile((entry) => {
  if (entry.url !== undefined) {
    if (entry.index === undefined) {
      errors.push(`${file}: Entity with URL '${entry.url}' should have an index.`);
    }

    resources[entry.url] = { index: entry.index, name: entry.name };
  }
});
```

and:
```js
forEachFile((entry) => {
  for (const property in entry) {
    if (typeof entry[property] === "object") {
      walk_recursive(resources, file, entry[property], errors);
    }
  }
});
```

src/tests/tables.test.js
```diff
@@ -20,3 +20,60 @@ describe("duplicate indices", () => {
...
+            errors.push(`${file}: URL '${object.url}' not found.`);
+          } else {
+            if (resources[object.url].name !== undefined && resources[object.url].name !== object.name) {
+              errors.push(`${file}: Name mismatch for reference to '${object.url}', '${object.name}' should be '${resources[object.url].name}'`);
+            }
+
+            if (object.index !== undefined && resources[object.url].index !== object.index) {
+              errors.push(`${file}: Index mismatch for reference to '${object.url}', '${object.index}' should be '${resources[object.url].index}'`);
+            }
+          }
```
bagelbits: You can probably just extract this out of the `for` loop since you just want to know if the object has the `url` key. Like this:
```js
if (object.hasOwnProperty('url') {
```

That will clean up your `for` loop to only be looking for subobjects.

src/tests/tables.test.js
```diff
@@ -20,3 +20,60 @@ describe("duplicate indices", () => {
...
 });
+
+describe("broken links", () => {
+  it("should not contain broken links", () => {
+    let errors = [];
+    let files = glob.sync("src/*.json");
+
+    let resources = {};
+
+    const walk_recursive = (resources, file, object, errors) => {
```
bagelbits: How do you feel about extracting this helper method outside of the tests?

bagelbits (commented):
Looks good! Just some cleanup

src/tests/tables.test.js
MikkelPaulson: Sure, that aligns well with a non-overlapping separation between `forEachFile()` and `forEachFileRecursive()`.

src/tests/tables.test.js
```diff
@@ -20,3 +20,60 @@ describe("duplicate indices", () => {
     expect(errors).toEqual([]);
   });
 });
+
+describe("broken links", () => {
```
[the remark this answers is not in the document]
MikkelPaulson: Went with `"api references"` for consistency with the other test case.

src/tests/tables.test.js
```diff
@@ -3,20 +3,94 @@ const glob = require("glob");
...
+ * Does not overlap with entries from forEachFile().
+ *
+ * @param (function(string, object)) callback Called with filename and each
+ *     sub-top-level entry.
+ */
+const forEachFileRecursive = (callback) => {
+  const recurse = (filename, entry) => {
+    for (const property in entry) {
+      if (typeof entry[property] === "object" && entry[property] !== null) {
+        callback(filename, entry[property]);
```
bagelbits: Should this callback be outside of the `for` loop? As it stands, it seems like it will skip the top level field? While that is what you're after, I think it limits the reuse of this method?

src/tests/tables.test.js
```diff
@@ -3,20 +3,94 @@ const glob = require("glob");
...
+      if (entry.url !== undefined) {
+        if (entry.index === undefined) {
+          errors.push(`${filename}: Entry with URL '${entry.url}' should have an index.`);
         }
-        indices.add(entry.index);
-      });
-    }
+
+        resources[entry.url] = { index: entry.index, name: entry.name };
+      }
```
bagelbits: Can we shift this to a guarded clause?

```js
if (entry.url === undefined) return;
if (entry.index === undefined) {
  errors.push(`${filename}: Entry with URL '${entry.url}' should have an index.`);
}

resources[entry.url] = { index: entry.index, name: entry.name };
```

src/tests/tables.test.js
```diff
@@ -3,20 +3,94 @@ const glob = require("glob");
...
+        } else {
+          if (resources[entry.url].name !== undefined && resources[entry.url].name !== entry.name) {
+            errors.push(`${filename}: Name mismatch for reference to '${entry.url}', '${entry.name}' should be '${resources[entry.url].name}'`);
+          }
+
+          if (entry.index !== undefined && resources[entry.url].index !== entry.index) {
+            errors.push(`${filename}: Index mismatch for reference to '${entry.url}', '${entry.index}' should be '${resources[entry.url].index}'`);
+          }
+        }
+      }
```
bagelbits: Can we also shift this to a guarded clause?

```suggestion
      if (!entry.hasOwnProperty("url")) return;
      
      if (resources[entry.url] === undefined) {
        errors.push(`${filename}: URL '${entry.url}' not found.`);
      } else {
        if (resources[entry.url].name !== undefined && resources[entry.url].name !== entry.name) {
          errors.push(`${filename}: Name mismatch for reference to '${entry.url}', '${entry.name}' should be '${resources[entry.url].name}'`);
        }

        if (entry.index !== undefined && resources[entry.url].index !== entry.index) {
          errors.push(`${filename}: Index mismatch for reference to '${entry.url}', '${entry.index}' should be '${resources[entry.url].index}'`);
        }
      }
```

src/tests/tables.test.js
```diff
@@ -3,20 +3,94 @@ const glob = require("glob");
...
+
+/**
+ * Calls the callback for all non-top-level objects/arrays in all JSON files.
+ * Does not overlap with entries from forEachFile().
+ *
+ * @param (function(string, object)) callback Called with filename and each
+ *     sub-top-level entry.
+ */
+const forEachFileRecursive = (callback) => {
+  const recurse = (filename, entry) => {
```
bagelbits: Is there a way to turn this into a function that will just recursively walk an object and take a callback? And then just feed that into the `forEachFile` as a callback? My brain is having a time trying to construct the code, but I feel like there is more reuse out of that function instead of tying it to `forEachFile`.

Thoughts on this?

src/tests/tables.test.js
```diff
@@ -3,20 +3,94 @@ const glob = require("glob");
 
 describe("duplicate indices", () => {
   it("should contain unique indices", () => {
+    let errors = [];
+    const fileIndices = {};
+
+    forEachFile((filename, entry) => {
```
bagelbits: Nice use of this!

bagelbits (commented):
Sorry. Some more comments but this feels like it's on the right track still.

src/tests/tables.test.js
MikkelPaulson: I think the general-case recursive walk function you're describing is exactly the one you're commenting on, just exposed at a broader scope.

src/tests/tables.test.js
MikkelPaulson: Maybe. My gut feeling is that having `forEachFile()` and `forEachFileRecursive()` as non-overlapping complements will be valuable, rather than having to check if you're dealing with a top-level entry or a reference.

MikkelPaulson (commented):
No problem at all, haven't had much in the way of code review since I became self-employed. It's nice to have someone else critique my JS for a change.

In general, I feel like further structural refactors can be filed under YAGNI and can be handled as necessary in the future.

src/tests/tables.test.js
bagelbits: Mainly, it feels confusing if the top level is skipped for the recursive? What I would expect `forEachFileRecursive` to run the callback on every level and not every non-top level?

MikkelPaulson: Fair. Rather than implementing that behaviour and then immediately working around it in the only place where the code is used, I've split out the recursive code as suggested and recombined it with `forEachFile()` in the new test. ~~`recurseInObject()`~~ `recurseIntoObject()` still has the same behaviour of not calling the callback with the top-level object, but I feel like that's more intuitive now that the object is coming directly from the calling code. I'm not married to that decision, though.

Maybe more problematic, it still only calls the callback for each *object* (or array), not other property types, which suits my use case just fine but might be less intuitive. Still, having your callback invoked with only `null` or `false` isn't very helpful without path information and/or the ability to mutate data in-place, so I'm inclined to let that use case lie and deal with it as needed.

src/tests/tables.test.js
```diff
@@ -3,20 +3,92 @@ const glob = require("glob");
...
+      if (entry.url === undefined) return;
+
+      if (entry.index === undefined) {
+        errors.push(`${filename}: Entry with URL '${entry.url}' should have an index.`);
+      }
+
+      resources[entry.url] = { index: entry.index, name: entry.name };
+    });
+
+    forEachFile((filename, topLevelEntry) => {
```
bagelbits: Last one I swear, thoughts on combining these?

```suggestion
    forEachFile((filename, topLevelEntry) => {
      if (topLevelEntry.url === undefined) return;

      if (topLevelEntry.index === undefined) {
        errors.push(`${filename}: Entry with URL '${topLevelEntry.url}' should have an index.`);
      }

      resources[topLevelEntry.url] = { index: topLevelEntry.index, name: topLevelEntry.name };
```

bagelbits (approved):
One small cleanup comment, but otherwise, good to go.

src/tests/tables.test.js
```diff
@@ -3,20 +3,92 @@ const glob = require("glob");
...
   });
 });
+
+/**
+ * Calls the callback for top-level objects/arrays in all JSON files.
+ *
+ * @param (function(string, object)) callback Called with filename and each
+ *     top-level entry.
+ */
+const forEachFile = (callback) => {
```
bagelbits: Would this be better named as `forEachEntryInFile`?

src/tests/tables.test.js
bagelbits: (It was not the last, so I guess I lied.)

bagelbits:
@MikkelPaulson At this point, it's just nitpicks.

src/tests/tables.test.js
MikkelPaulson: No, we have to populate all defined URLs into `resources` before we can begin the validation pass.

src/tests/tables.test.js
MikkelPaulson: Makes sense, but I went with `forEachFileEntry()` instead because it iterates over all files.

src/tests/tables.test.js
bagelbits: My bad! I totally missed that.

Issue #349, opened 2021-02-18
Update hand crossbows proficiency name to be consistent with proficiency lists in the SRD.


fergcb:
In the SRD, proficiencies for bards and rogues include hand crossbows as "hand crossbows". In the SRD, the hand crossbow equipment item is listed as "Crossbow, hand".

In the data, the Proficiency document for hand crossbows, `crossbows-hand`, has the name "Crossbows, hand". This is consistent with the name of the item, but not with how the SRD lists the actual proficiency. The `name` (and `index`) should be updated to "hand crossbows" (and "hand-crossbows") to better reflect the way the SRD defines the proficiency.

PR #402, opened 2021-08-04
Update hand crossbows proficiency name to be consistent with proficiency lists in the SRD. #349

rbmy:
## What does this do?
Update hand crossbows proficiency name to be consistent with proficiency lists in the SRD. #349

## How was it tested?
CI

#349 

random photo - update me


bagelbits (approved):
LGTM! Thanks for doing this!

PR #406, opened 2021-08-11
Change id on monster forms to index

SleeplessOne1917:
## What does this do?
I noticed that the API references for monster forms used id instead of index. This seems inconsistent with how other API references are handled, so I changed it to index.

## How was it tested?
I ran dbRefresh, opened MongoDB Compass, and viewed all the entries with forms defined to make sure the change went through

No

n/a

random photo - update me

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -26202,9 +26202,7 @@
         }
       }
     ],
-    "damage_vulnerabilities": [
-      "fire"
-    ],
+    "damage_vulnerabilities": ["fire"],
     "damage_resistances": [],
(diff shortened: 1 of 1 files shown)


src/5e-SRD-Monsters.json
```diff
@@ -26202,9 +26202,7 @@
         }
       }
     ],
-    "damage_vulnerabilities": [
-      "fire"
-    ],
+    "damage_vulnerabilities": ["fire"],
```
SleeplessOne1917, line 26205: Prettier's autoformat on save got this one.

bagelbits (approved):
Good catch!

PR #411, opened 2021-08-17
Make heavy crossbow heavy

SleeplessOne1917:
## What does this do?
Changes heavy crossbow to be heavy instead of light.

## How was it tested?
I ran dbRefresh and checked MongoDB compass to see that the property was updated.

No

n/a

random photo - update me


bagelbits (approved):
Good catch!

PR #414, opened 2021-08-19
Proficiency reference breaking

### LINKED ISSUE #413: Proficiencies: Return object instead of single item array
I was looking through the proficiencies collection and noticed that for every proficiency, `references` is always an array with 1 entry. Yes, I checked them all. Considering the API used to reach this database only supports reads, we don't have to worry about inserting a proficiency with multiple references.

I propose we change `references`  to `reference` and that we nix the array.

SleeplessOne1917:
## What does this do?
Changes `references` in proficiencies to `reference`, return object instead of array of single object. Remove `type` from reference.

## How was it tested?
I verified the changes in MongoDB compass. I also ran the integration tests on the API project.

This PR resolves #413 

[Yes](https://github.com/5e-bits/5e-srd-api/pull/242)

random photo - update me

### DIFF
--- a/src/5e-SRD-Proficiencies.json
+++ b/src/5e-SRD-Proficiencies.json
@@ -42,14 +42,12 @@
     ],
     "races": [],
     "url": "/api/proficiencies/light-armor",
-    "references": [
-      {
-        "index": "light-armor",
-        "type": "equipment-categories",
-        "name": "Light Armor",
-        "url": "/api/equipment-categories/light-armor"
-      }
-    ]
+    "reference": {
+      "index": "light-armor",
+      "type": "equipment-categories",
+      "name": "Light Armor",
+      "url": "/api/equipment-categories/light-armor"
+    }
   },
   {
     "index": "medium-armor",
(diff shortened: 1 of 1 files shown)


bagelbits (commented):
This actually does require an update to the documentation on the API in [this file](https://github.com/5e-bits/5e-srd-api/blob/main/src/views/partials/doc-resource-proficiencies.ejs)

You can find it on the docs here: http://www.dnd5eapi.co/docs/#proficiencies

bagelbits:
I think other than that, this solidly handles #413

bagelbits (dismissed):
LGTM!

bagelbits:
@abias1122 Due to this being a breaking change, can we split this into two PRs? Namely, one PR for adding the `reference` key/value and another for removing the `references` key/value? That will let me merge the addition immediately and give everyone a week to change over before removing `references`? I'll merge in your API doc update with the first PR.

SleeplessOne1917:
@bagelbits The non-breaking version of this PR can be found [here](https://github.com/5e-bits/5e-database/pull/416).

bagelbits:
@abias1122 There are some conflicts to resolve but then this should be good to merge in.

bagelbits (approved):
Once we resolve conflicts, this is good to go.

Issue #337, opened 2021-01-06
Use Markdown to represent rich text


fergcb:
# The Problem
Currently, there are some oddities with the way the API represents rich text fields, such as spell descriptions. The present strategy is to separate paragraphs into their own strings, as items in an array. This is practical for rendering - `for each` over the array, and produce a `<p>` tag for each string (or your stack's equivalent); however, this approach is severely limiting when it comes to complex mark-up, such as tables, subheadings, etc.

Compliance with this approach leads to some interesting formatting in the API's data. For example, the spell `scrying` has two tables, in the source:

![image](https://user-images.githubusercontent.com/6972523/103794027-21c8ce00-503c-11eb-93cf-097cdd5f4dfc.png)

Whereas the format given in the API is distant from the source, and carries no information about emphasis, where tables begin and end, and treats table rows as distinct paragraphs:

```json
{
  "index": "scrying",
  "name": "Scrying",
  "desc": [
    "You can see and hear a particular creature you choose that is on the same plane of existence as you. The target must make a wisdom saving throw, which is modified by how well you know the target and the sort of physical connection you have to it. If a target knows you're casting this spell, it can fail the saving throw voluntarily if it wants to be observed.",
    "Knowledge & Save Modifier",
    "Secondhand (you have heard of the target) +5",
    "Firsthand (you have met the target) +0",
    "Familiar (you know the target well) -5",
    "Connection & Save Modifier",
    "Likeness or picture -2",
    "Possession or garment -4",
    "Body part, lock of hair, bit of nail, or the like -10",
    "On a successful save, the target isn't affected, and you can't use this spell against it again for 24 hours.",
    "On a failed save, the spell creates an invisible sensor within 10 feet of the target. You can see and hear through the sensor as if you were there. The sensor moves with the target, remaining within 10 feet of it for the duration. A creature that can see invisible objects sees the sensor as a luminous orb about the size of your fist.",
    "Instead of targeting a creature, you can choose a location you have seen before as the target of this spell. When you do, the sensor appears at that location and doesn't move."
  ],
  "...": "..."
}
```

# Proposed Solution: Markdown
[...]

PR #415, opened 2021-08-20
Clean magic item descriptions

MikkelPaulson:
## What does this do?

* Updates formatting of Deck of Many Things (original source material includes some items denoted with `*`)
* Removes the description of sentient magic items that somehow got lumped in with Wings of Flying
* Removes Potion of Healing from Equipment.json
* Adds formatting from the source material to magic item descriptions, eg:

> The typical crystal ball, a very rare item, is about 6 inches in diameter. While touching it, you can cast the scrying spell (save DC 17) with it.
> The following crystal ball variants are legendary items and have additional properties.
> ***Crystal Ball of Mind Reading.*** You can use an action to cast the detect thoughts spell (save DC 17) while you are scrying with the crystal ball, targeting creatures you can see within 30 feet of the spell's sensor. You don't need to concentrate on this detect thoughts to maintain it during its duration, but it ends if scrying ends.
> ***Crystal Ball of Telepathy.*** While scrying with the crystal ball, you can communicate telepathically with creatures you can see within 30 feet of the spell's sensor. You can also use an action to cast the suggestion spell (save DC 17) through the sensor on one of those creatures. You don't need to concentrate on this suggestion to maintain it during its duration, but it ends if scrying ends. Once used, the suggestion power of the crystal ball can't be used again until the next dawn.
> ***Crystal Ball of True Seeing.*** While scrying with the crystal ball, you have truesight with a radius of 120 feet centered on the spell's sensor.

## How was it tested?

* Ran jest tests (which caught a spot where I forgot to remove the reference to potion of healing from another file, so tests work!)
* Consuming the data set myself - haven't tested Docker this time

Might be peripherally related to #337.

N/A


bagelbits (approved):
Thanks for the cleanup! Add in the Markdown will make this even nicer!

PR #418, opened 2021-08-22
Remove attack_bonus from parry

SleeplessOne1917:
## What does this do?
Removed attack bonus from gladiator's parry since it made no sense.

## How was it tested?
Ran db:refresh, checked in MongoDB compass

No

n/a

random photo - update me


bagelbits (approved):
Yeah. That doesn't make any sense.

PR #419, opened 2021-08-22
Remove attack bonus from relentless

SleeplessOne1917:
## What does this do?
`attack_bonus` didn't seem applicable to the relentless ability, so I removed it.

## How was it tested?
Ran db:refresh and verified changes in MongoDB compass

No

n/a

random photo - update me

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -37438,8 +37438,7 @@
         "usage": {
           "type": "recharge after rest",
           "rest_types": ["short", "long"]
-        },
-        "attack_bonus": 0
+        }
       }
     ],
     "actions": [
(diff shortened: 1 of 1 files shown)


src/5e-SRD-Monsters.json
```diff
@@ -37525,8 +37524,7 @@
         "usage": {
           "type": "recharge after rest",
           "rest_types": ["short", "long"]
-        },
-        "attack_bonus": 0
```
SleeplessOne1917: For some reason the diff shows this change being under actions, which is incorrect. The change was actually under special abilities.

bagelbits: It isn't actually showing that. It's just that Github doesn't know how to handle really big files for expansion.

bagelbits (approved):
LGTM!

PR #421, opened 2021-08-27
Add extra language trait to high elf

SleeplessOne1917:
## What does this do?
I noticed that the subraces field of the extra language trait referenced high elf, but not the other way around. This PR fixes that.

## How was it tested?

no

n/a

random photo - update me


bagelbits (approved):
Good catch!

PR #422, opened 2021-09-04
Add spell save dc and attack modifier to spellcasting ability info

SleeplessOne1917:
## What does this do?
I noticed that the spellcasting info for bard and cleric included calculations for spell save dc and spell attack modifier, but the other spellcasting classes lacked this information. This PR makes this consistent, adding the previously mentioned information to all spellcasting classes.

## How was it tested?

n/a

n/a

random photo - update me


bagelbits (approved):
Thanks for adding this!

PR #425, opened 2021-09-05
Update prerequisites for fiend subclass spells

### LINKED ISSUE #424: Fiend subclass spell prerequisites seem off
While looking at the subclass spells, I noticed that all of the prerequisites for the fiend subclass reference warlock level 1. This doesn't seem correct. I'll be using information from the [Roll20 Compendium page on the warlock](https://roll20.net/compendium/dnd5e/Classes:Warlock?expansion=0#h-The%20Fiend) to show you what I mean.

The table for the fiend spells is the following:
|Spell Levels | Spells |
|-------------|--------|
| 1st| Burning Hands, Command|
| 2nd| Blindness/Deafness, Scorching Ray|
| 3rd| Fireball, Stinking Cloud|
| 4th| Fire Shield, Wall of Fire|
| 5th | Flame Strike, Hallow |

Also worth pointing out is the compendium's entry on spells known of first level or higher.

> You learn a new Warlock spell every time you gain a level from 2 through 9, as well as at level 19. **A spell you choose must be of a level no higher than what's shown in the table's Slot Level column for your level.** When you reach 6th level, for example, you learn a new Warlock spell, which can be 1st, 2nd, or 3rd Level.

Looking at these two pieces of information, it doesn't seem to make sense making all of the fiend pact's spells prerequisite's level 1. That would mean, for example, that a player could use Flame Strike, a 5th level spell, when they only have 1st level spell slots available.

I propose that the prerequisite levels be changed to match when the warlock will have slots for a given spell.

SleeplessOne1917:
## What does this do?
Updates the warlock's pact of the fiend subclass to have spell prerequisites match when the warlock will get spell slots for them.

## How was it tested?

This PR resolves #424 

n/a

random photo - update me

### DIFF
--- a/src/5e-SRD-Subclasses.json
+++ b/src/5e-SRD-Subclasses.json
@@ -1662,10 +1662,10 @@
       {
         "prerequisites": [
           {
-            "index": "warlock-1",
+            "index": "warlock-3",
             "type": "level",
-            "name": "Warlock 1",
-            "url": "/api/classes/warlock/levels/1"
+            "name": "Warlock 3",
+            "url": "/api/classes/warlock/levels/3"
           }
         ],
         "spell": {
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
LGTM!

Issue #427, opened 2021-10-09
Inconsistencies in damage data from monsters


TheLazyDutchman:
I saw that on of the actions in the Ancient Blue Dragon had this structure in the api
```
{
	"name": "Lightning Breath",
	"desc": "The dragon exhales lightning in a 120-foot line that is 10 feet wide. Each creature in that line must make a DC 23 Dexterity saving throw, taking 88 (16d10) lightning damage on a failed save, or half as much damage on a successful one.",
	"usage": {
		"type": "recharge on roll",
		"dice": "1d6",
		"min_value": 5
	},
	"dc": {
		"dc_type": {
			"index": "dex",
			"name": "DEX",
			"url": "/api/ability-scores/dex"
		},
		"dc_value": 23,
		"success_type": "half"
	},
	"damage_dice": "16d10",
	"damage": []
}
```

as you can see, it has a "damage_dice" attribute outside of an empty "damage" list.
As far as I know, all other actions follow the following pattern for damage.

```
{
	"name": "Tail",
	"desc": "Melee Weapon Attack: +16 to hit, reach 20 ft., one target. Hit: 18 (2d8 + 9) bludgeoning damage.",
	"attack_bonus": 16,
	"damage": [
		{
			"damage_type": {
				"index": "bludgeoning",
				"name": "Bludgeoning",
				"url": "/api/damage-types/bludgeoning"
			},
			"damage_dice": "2d8+9"
		}
	]
}
```

I don't know if this is something in the data itself, or if it comes from the API.
nor do I know if there are any other places where "damage_dice" is outside the "damage" list.

I think that for consistency's sake the "damage_dice" should be placed inside a damage object, probably with a "damage_type" of lightning, since that is consistent with the statblock on dndbeyond (and the fact that, you know, the action is called "Lightning Breath".

bagelbits:
That's an issue in the database. I think that just got missed when I switched over the shape. If you want to open a quick PR, please feel free to fix it.

PR #428, opened 2021-10-13
fixed inconsistencies for ancient blue dragon

TheLazyDutchman:
## What does this do?
the lightning breath for ancient dragon had no "damage" list, just a separate "damage_dice" property
this should fix that

## How was it tested?
I have not tested this change

#427 

random photo - update me


bagelbits (approved):
Thank you for catching this and fixing it!

PR #429, opened 2021-10-26
Fix Detect Thoughts description

MikkelPaulson:
## What does this do?
Update the description of the Detect Thoughts spell to match SRD 5.1. Not sure where the previous content came from, but it was... wrong.

## How was it tested?
Jest passed. Otherwise, this is a content-only change.

No.

N/A

random photo - update me


bagelbits (approved):
Good catch!

PR #432, opened 2021-11-08
Fix damage and condition immunities for some undead

SleeplessOne1917:
## What does this do?
Some of the damage and condition immunities for undead monsters were wrong, so I fixed them.

## How was it tested?
I ran db:refresh and verified the changes in MongoDB compass.

No

n/a

random photo - update me


bagelbits (approved):
Good catch!

PR #433, opened 2021-12-07
I'm pretty sure that's not needed

bagelbits:
## What does this do?

Removes a piece of the CI because it's probably not necessary.

## How was it tested?

One of those, we gotta do it live tests.

Nope

N/A

image


.github/workflows/ci.yml
```diff
@@ -52,9 +52,6 @@ jobs:
     needs: deploy
     if: github.ref == 'refs/heads/main'
```
bagelbits, line 53:
```suggestion
```

PR #434, opened 2021-12-12
Feature localization

Platiplus:
## What does this do?

Proof of concept about localization of the database

## How was it tested?

run `node ./scripts/createLanguageFiles.js`

#81 

\<It's not clear if I don't update this text with relevant info\>

random photo - update me


src/5e-SRD-Weapon-Properties.json
```diff
@@ -1,88 +1,90 @@
 [
   {
     "index": "ammunition",
-    "name": "Ammunition",
+    "name": "ammunitionName",
     "desc": [
-      "You can use a weapon that has the ammunition property to make a ranged attack only if you have ammunition to fire from the weapon. Each time you attack with the weapon, you expend one piece of ammunition. Drawing the ammunition from a quiver, case, or other container is part of the attack (you need a free hand to load a one-handed weapon).",
-      "At the end of the battle, you can recover half your expended ammunition by taking a minute to search the battlefield. If you use a weapon that has the ammunition property to make a melee attack, you treat the weapon as an improvised weapon (see \"Improvised Weapons\" later in the section). A sling must be loaded to deal any damage when used in this way."
+      "ammunitionDesc1",
+      "ammunitionDesc2"
```
bagelbits, line 7: Would it make more sense for `ammunitionDesc` to be a list instead?

localization/keys/en-us/5e-SRD-Weapon-Properties.json
```diff
@@ -0,0 +1,25 @@
...
+  "ammunitionDesc2": "At the end of the battle, you can recover half your expended ammunition by taking a minute to search the battlefield. If you use a weapon that has the ammunition property to make a melee attack, you treat the weapon as an improvised weapon (see \"Improvised Weapons\" later in the section). A sling must be loaded to deal any damage when used in this way." ,
+  "finesseName": "Finesse",
+  "finesseDesc1": "When making an attack with a finesse weapon, you use your choice of your Strength or Dexterity modifier for the attack and damage rolls. You must use the same modifier for both rolls.",
+  "heavyName": "Heavy",
+  "heavyDesc1": "Small creatures have disadvantage on attack rolls with heavy weapons. A heavy weapon's size and bulk make it too large for a Small creature to use effectively.",
+  "lighName": "Light",
```
bagelbits, line 9: Should this be `lightName` instead of `lighName`?

localization/en-us/.gitignore
```diff
@@ -0,0 +1 @@
+*.json
```
bagelbits, line 1: Why did we add these to the git ignores?

bagelbits (commented):
A couple of comments to start. I'm looking forward to talking to you about this more tomorrow.

src/5e-SRD-Weapon-Properties.json
Platiplus: Yes and No. That's one of my concerns about this approach instead of just mimicking the whole file structure and start translating it. We'll talk more today.

localization/keys/en-us/5e-SRD-Weapon-Properties.json
Platiplus: Yes. Freaking typo haha

localization/en-us/.gitignore
Platiplus: Just added this gitignore so we would't commit the translated files by accident when messing around with the script. Also PERSONALLY I think that this keeps the folder created for us on git, so we shouldn't fall on some edge cases when we have to account for cache and checking if the file is already there or not in the script, I had some problems with this in the past, but is totally optional.

PR #435, opened 2021-12-22
Correct wording of spells containing "Fate"

MikkelPaulson:
## What does this do?

In response to user feedback on initiative.sh, corrected the wording of the Legend Lore spells. A cursory check revealed a number of other spells, all containing the word "fate", with strange deviations from the SRD. They read like someone was attempting to translate from another language.

Full list of affected spells:

* Glyph of Warding
* Legend Lore
* Mirror Image
* Spike Growth
* Zone of Truth

Also removed non-ASCII characters from Control Weather and Earthquake,
for consistency's sake ("×" => "x").

## How was it tested?

Rebuilt and tested app with updated data.

No.

N/A

really-mom-i-put-sun-screen-on-already

### DIFF
--- a/src/5e-SRD-Spells.json
+++ b/src/5e-SRD-Spells.json
@@ -3151,7 +3151,7 @@
     "name": "Control Weather",
     "desc": [
       "You take control of the weather within 5 miles of you for the duration. You must be outdoors to cast this spell. Moving to a place where you don't have a clear path to the sky ends the spell early.",
-      "When you cast the spell, you change the current weather conditions, which are determined by the DM based on the climate and season. You can change precipitation, temperature, and wind. It takes 1d4 × 10 minutes for the new conditions to take effect. Once they do so, you can change the conditions again. When the spell ends, the weather gradually returns to normal.",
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
LGTM!

ogregoire:
I disagree with the change from `×` to `x`: the SRD uses the first and not the latter. See "Control Weather" on page 131 of the SRD for proof.

The SRD doesn't bother with being ASCII-compliant, so why should we?

bagelbits:
@ogregoire I could swear that we've been moving away from non-ASCII for a while now.

Granted, if we ever fully support localization, that might not make any sense anymore.

MikkelPaulson:
Those two uses of "×" were the only matches in the file for /[^\x00-\x7f]/. All other cases (if there are any) were already "x".

The SRD also uses ‘curly quotes’, which is why I went looking for non-ASCII characters in the first place. (I copy-pasted the correct descriptions out of the PDF and reformatted as necessary.)

I don't have a strong opinion either way, but if you want to unify behind ASCII characters or especially if you want to unify behind non-ASCII punctuation, that should be enforced by tests.

ogregoire:
Unicode allows us to change curly quotes into straight ones. The multiplication sign into an 'x' isn't.

bagelbits:
@ogregoire I'm going with us using non-ASCII where it makes sense as long as we're consistent.

PR #439, opened 2022-01-06
Simplify Markdown's stars when possible

ogregoire:
## What does this do?

Simplify the use of stars from the hard to read/understand `\\*` to the easier `*` when possible.

## How was it tested?

Tested locally in a Markdown interpreter

No, but it's discussed here: https://github.com/5e-bits/5e-database/commit/86ef38e72ae955d9372ba9a97315fba58fe26836

No

random photo - update me


src/5e-SRD-Magic-Items.json
```diff
@@ -947,27 +947,27 @@
...
+      "| Ace of spades | Donjon* |",
       "| King of spades | Ruin |",
       "| Queen of spades | Euryale |",
       "| Jack of spades | Rogue |",
-      "| Two of spades | Balance\\* |",
-      "| Joker (with TM) | Fool\\* |",
+      "| Two of spades | Balance* |",
+      "| Joker (with TM) | Fool* |",
       "| Joker (without TM) | Jester |",
       "\\* Found only in a deck with twenty-two cards",
```
bagelbits:
```suggestion
      "* Found only in a deck with twenty-two cards",
```

bagelbits (approved):
I think this is one of those syntax style things. Both of these are correct, but the initial was being cautious because `*` is a special character in Markdown.

PR #440, opened 2022-01-13
Fixed a minor typo in Monsters JSON

megoth:
## What does this do?

Fixes a minor typo in Monsters JSON

## How was it tested?

Ran `npm run lint` and `npm run test`

No, it's a simple typo, so didn't think it necessary.

Don't think it's necessary

here you go


bagelbits (approved):
Good catch!

PR #445, opened 2022-02-24
perf(multilingual-support): Multi-language Support

djurnamn:
## What does this do?

This is another approach to adding multilingual support to 5e-database. It allows us to parse the source data and separate it into what should be translatable (locale) and what shouldn't be (templates). It also allows us to build the source data back together with an altered locale file, resulting in a translated version of the database. Please see the [readme](https://github.com/djurnamn/5e-database/blob/multilingual-support/multilingual-support/readme.md) for more information!

## How was it tested?

I have manually tested to take the original data apart and put it back together again using the scripts. The result almost perfectly replicates the original data. There's just one instance of a comma separated list in the original data where the `and` is second to last instead of last, which is the (otherwise reliable) pattern I've set up. I also wasn't able to perfectly replicate the json formatting options of the original data, so I made some changes to the source data (mainly indentation and newlines) to be able to compare the results more reliably.

#81 

I did have a look at the proof of concept that @carloslancha made a while back (#158). That helped me get started!

I've added a [readme](https://github.com/djurnamn/5e-database/blob/multilingual-support/multilingual-support/readme.md) in the subfolder. But this is pretty self-contained and non-intrusive. It doesn't affect how the API works.

random photo - update me


djurnamn:
I did some more experimenting on this. It's not super pretty, but it seems to do the trick. The easiest way to test it is to:
1.  run `npm run create-source-locale` which will now create `locales/en_us.json`
2. duplicate it and rename the copy to some other country code (and maybe change some of the values in the file as well)
3. run `npm run populate-templates` which will now create a set of source files containing both languages in the desired shape (in the `output` folder)

I'll try to do some refactoring and documentation on this when I have more time (as it's not super straight forward). But feel free to test it out and let me know what you think!

PR #448, opened 2022-03-10
ecs: fix deploy ci

ecshreve:
## What does this do?
I noticed the last two deploys failed, looks to be caused by types not being accessible when we try to execute `npm run db:refresh`.

Updates the `Deploy` job in the CI workflow to run `npm install` before `npm run db:refresh`.

## How was it tested?
- Confirmed I see the same errors locally when I attempt `npm run db:refresh` with a fresh copy of the repo. Running `npm install` resolves the errors.

na

na

random photo - update me

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,9 +43,12 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 14.x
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
LGTM

PR #449, opened 2022-03-11
Fix shriekers reaction from action

DrZoddiak:
## What does this do?

Quick change to fix Shriekers action to actually be a reaction

## How was it tested?

It was not.

no

no

image


bagelbits (approved):
Thanks!

PR #451, opened 2022-03-24
feat: release docker image to github container registry

brittonhayes:
## What does this do?

- Releases the docker image to the github container registry upon tagging of a new version


bagelbits (approved):
Neat! I'll have to start to use Semantic Versioning but I think that's good. It'll be extra nice since I need to close out a decision with multi-language support.

Issue #367, opened 2021-05-21
Improvement - Extract Magic Item data from Description


mjl757:
There is data in the description field of a magic items that would be useful to extract out into their own fields for better access.
The first item in the description array holds the item type, as well as the the rarity (or rarities in some instances). So my proposal would be to add a couple fields to the magic item.

First one would be an object named magic_item_category:
```
{
  "name":  __,
  "desc": __
}
```

And the second one would be an array named rarity:

```
[
  {
    "name": __,
     "desc":  __
  }
]
```

Example: 

```
{
  "index": "ammunition",
  "name": "Ammunition, +1, +2, or +3",
  "equipment_category": {
    "index": "ammunition",
    "name": "Ammunition",
    "url": "/api/equipment-categories/ammunition"
  },
  "magic_item_category": {
    "name": "Weapon",
    "desc": "any ammunition"
  },
  "rarity": [
    {
      "name": "uncommon",
      "desc": "+1"
    },
    {
      "name": "rare",
      "desc": "+2"
    },
    {
      "name": "very rare",
      "desc": "+3"
    }
  ],
  "desc": [
    "Weapon (any ammunition), uncommon (+1), rare (+2), or very rare (+3)",
    "You have a bonus to attack and damage rolls made with this piece of magic ammunition. The bonus is determined by the rarity of the ammunition. Once it hits a target, the ammunition is no longer magical."
  ],
  "url": "/api/magic-items/ammunition"
}
```

Certainly interested in having a discussion on these purposed changes and seeing what we can work out.

mjl757:
One possible issue see with it is that the category isn't quite in line with other category fields, in that it's not indexable. with it's own table and everything. So I could see us making that more inline with gear and equipment categories then adding a third field to magic_items `category_description` which would be the description field in the category. 

So the above example would become something like this:

```
{
 ...
  "magic_item_category": {
      "index": "weapon
      "name": "Weapon",
     "url": "/api/magic-item-categories/weapon"
    },
  "magic_item_category_desc": "any ammunition", 
  "rarity": [
    {
      "name": "uncommon",
      "desc": "+1"
    },
    {
      "name": "rare",
      "desc": "+2"
    },
    {
      "name": "very rare",
      "desc": "+3"
    }
  ],
...
}
```

bagelbits:
[...]

PR #455, opened 2022-03-26
feat(magic-items): add rarity to magic items

ivorisoutdoors:
## What does this do?

This splits out item variants, creating an entry for each variant of a magic item and text to match what the variant does. The original item creates a link to each variant and keeps the original item text.

This also adds rarity to each item. In cases where there are variants of different rarity, the value is "Varies".

## How was it tested?

Generated a simple json schema and made sure all items had a rarity

addresses some of #367 

5e-bits/5e-srd-api#293

explorer


src/5e-SRD-Magic-Items.json
```diff
@@ -21,12 +25,86 @@
...
     "desc": [
       "Weapon (any ammunition), uncommon (+1), rare (+2), or very rare (+3)",
       "You have a bonus to attack and damage rolls made with this piece of magic ammunition. The bonus is determined by the rarity of the ammunition. Once it hits a target, the ammunition is no longer magical."
     ],
     "url": "/api/magic-items/ammunition"
   },
+  {
+    "index": "ammunition-1",
+    "name": "Ammunition, +1",
+    "equipment_category": {
```
bagelbits, line 59: We'll need to updated [Equipment Categories](https://github.com/5e-bits/5e-database/blob/main/src/5e-SRD-Equipment-Categories.json) as well.

src/5e-SRD-Magic-Items.json
```diff
@@ -330,6 +554,102 @@
...
+      "name": "Wondrous Items",
+      "url": "/api/equipment-categories/wondrous-items"
+    },
+    "rarity": {
+      "name": "Uncommon"
+    },
+    "variants": [],
+    "desc": [
+      "Wondrous item, uncommon",
+      "This ordinary bag, made from gray, rust, or tan cloth, appears empty. Reaching inside the bag, however, reveals the presence of a small, fuzzy object. The bag weighs 1/2 pound.",
```
bagelbits:
```suggestion
      "This ordinary bag, made from tan cloth, appears empty. Reaching inside the bag, however, reveals the presence of a small, fuzzy object. The bag weighs 1/2 pound.",
```

src/5e-SRD-Magic-Items.json
```diff
@@ -21,12 +25,86 @@
...
+      }
+    ],
     "desc": [
       "Weapon (any ammunition), uncommon (+1), rare (+2), or very rare (+3)",
       "You have a bonus to attack and damage rolls made with this piece of magic ammunition. The bonus is determined by the rarity of the ammunition. Once it hits a target, the ammunition is no longer magical."
     ],
     "url": "/api/magic-items/ammunition"
   },
+  {
+    "index": "ammunition-1",
```
mjl757, line 57: What are your thoughts on putting the variants into their own json file. 

Could be something like `5e-SRD-Magic-Item-Varients.json` ? That would allow this table to easily still be used to pull up items as they are in the DMG, while giving people the flexibility to pull the individual variants up. 

As it stands the only way to filter out the variants is to pull the whole list then the consumer would have to loop over all the items and find out what to remove on their end.

bagelbits (commented):
I think this is a great start! Just two comments.

src/5e-SRD-Magic-Items.json
mjl757: It's certainly doable to keep everything in this file, I could just see it being a bit more complicated for people who are using it. I don't think it's a deal breaker either way though.

bagelbits: So, on the one hand, I can see the value in splitting that out like that. This also allows the api to do: `/api/magic-items/ammunition/variants` to give you only the variants.

On the other hand, it becomes a lot hard to get all the magic items of a certain rarity.

mjl757: Does it become that much harder? Couldn't you just query both tables and combine the results?

bagelbits: You're right. It doesn't become that much harder. But it feels like it's shifting the problem?

How about this as an alternative? What if you added a `variant: Boolean` field into all the entries. This would allow you keep it together but also allow you to easily query for things that aren't variants?

mjl757: I think that's a fine alternative. I was going to suggest that, but for some reason I had it in my head that flags like that weren't popular in this project. I don't know why.

bagelbits: Flags work when they make sense.

src/5e-SRD-Magic-Items.json
```diff
@@ -2138,419 +4485,973 @@
...
+      "Potion, uncommon",
+      "When you drink this potion, you gain the \"enlarge\" effect of the enlarge/reduce spell for 1d4 hours (no concentration required). The red in the potion's liquid continuously expands from a tiny bead to color the clear liquid around it and then contracts. Shaking the bottle fails to interrupt this process."
     ],
-    "url": "/api/magic-items/periapt-of-wound-closure"
+    "url": "/api/magic-items/potion-of-growth"
   },
   {
-    "index": "philter-of-love",
-    "name": "Philter of Love",
+    "index": "potion-of-healing",
```
ivorisoutdoors, line 5248: How do we want to handle this item? This is a case where the variant would have the same name as the parent item

ivorisoutdoors: For now, I added an item with the same name, but with an id of `potion-of-healing-common`

src/5e-SRD-Magic-Items.json
ivorisoutdoors: updated

src/5e-SRD-Magic-Items.json
ivorisoutdoors: added a variant flag to each item

src/5e-SRD-Magic-Items.json
bagelbits: I love how the SRD always has a weird curve ball for everything.

I think that's reasonable for now and we can adjust later.

bagelbits (approved):
Thanks for doing this! I'm sure we'll continue to iterate on this since Magic Items are in Beta still.

bagelbits:
WHOOPS. I merged this too early.

PR #456, opened 2022-03-29
feat/auto-release

brittonhayes:
## What does this do?

Adds automated semantic release github action.


.github/workflows/github-release.yml
```diff
@@ -0,0 +1,26 @@
...
+        - name: Setup Node.js
+          uses: actions/setup-node@v2
+          with:
+            node-version: 'lts/*'
+        - name: Install dependencies
+          run: npm install
+        - name: Release
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          run: npx semantic-release
```
bagelbits:
```suggestion
          run: npx semantic-release

```

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,41 @@
+name: container-release
```
bagelbits: In one of these release scripts, should we also run the release script that populates the database?

Also should this replace the other release flow?

bagelbits (approved):
A couple of questions but otherwise looks solid.

.github/workflows/container-release.yml
brittonhayes: Yeah I think that's a good idea. Just get it all goin in one place.

PR #491, opened 2022-07-29
fix(saving throw): Add saving throw proficiencies to class proficiencies

SleeplessOne1917:
## What does this do?

Adds saving throws to class proficiencies, as well as adding light crossbow proficiency to wizard and sorcerer.

## How was it tested?

I built the API with the database changes and queried the API.

n/a

n/a

random photo - update me


bagelbits (approved):
Good catch!

PR #495, opened 2022-08-07
feat: Add size and shape to dragonborn's breath weapon

SleeplessOne1917:
## What does this do?

Adds size and shape of dragonborn's breath weapon since the API didn't have that information already

## How was it tested?

I ran `db:refresh` and looked at the data with mongo db compass.

n/a

[Yes](https://github.com/5e-bits/5e-srd-api/pull/322).

random photo - update me

### DIFF
--- a/src/5e-SRD-Traits.json
+++ b/src/5e-SRD-Traits.json
@@ -759,6 +759,7 @@
       "breath_weapon": {
         "name": "Breath Weapon",
         "desc": "You can use your action to exhale destructive energy. Your draconic ancestry determines the size, shape, and damage type of the exhalation. When you use your breath weapon, each creature in the area of the exhalation must make a saving throw, the type of which is determined by your draconic ancestry. The DC for this saving throw equals 8 + your Constitution modifier + your proficiency bonus. A creature takes 2d6 damage on a failed save, and half as much damage on a successful one. The damage increases to 3d6 at 6th level, 4d6 at 11th level, and 5d6 at 16th level. After you use your breath weapon, you can't use it again until you complete a short or long rest.",
+        "size_and_shape": "5 by 30 ft. line",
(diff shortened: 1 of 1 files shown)


src/5e-SRD-Traits.json
```diff
@@ -759,6 +759,7 @@
       "breath_weapon": {
         "name": "Breath Weapon",
         "desc": "You can use your action to exhale destructive energy. Your draconic ancestry determines the size, shape, and damage type of the exhalation. When you use your breath weapon, each creature in the area of the exhalation must make a saving throw, the type of which is determined by your draconic ancestry. The DC for this saving throw equals 8 + your Constitution modifier + your proficiency bonus. A creature takes 2d6 damage on a failed save, and half as much damage on a successful one. The damage increases to 3d6 at 6th level, 4d6 at 11th level, and 5d6 at 16th level. After you use your breath weapon, you can't use it again until you complete a short or long rest.",
+        "size_and_shape": "5 by 30 ft. line",
```
bagelbits: Would it make sense to split this up more? Since we can standardize shape and use it in other places (like spells). And we can probably standardize size too.

SleeplessOne1917: Something like:
```
{
    size: string;
    shape: string;
}
```
?

bagelbits: Actually, I think Spells already have an `area_of_effect` field. It would probably make more sense to use that and potentially expand on it? Just for consistency. I can also see this being used for some monster attacks.

SleeplessOne1917: image
So something like this then?

bagelbits: Yeah.

bagelbits (approved):

PR #498, opened 2022-08-19
feat: Redo Warlock Eldritch Invocations

SleeplessOne1917:
## What does this do?

Originally in the API, Eldritch Invocations were split out into 8 different features with a choice with `choose` 2 from an incomplete list of invocations. This change puts them all into one feature with `choose` set to a string value for invocations known, since that's what will really be determining how many the warlock can learn. I also made sure the list of invocations was complete per the [SRD](https://media.wizards.com/2016/downloads/DND/SRD-OGL_V5.1.pdf).

## How was it tested?

I ran `db:refresh` and checked that the data was as expected. I also ran the API and queried it to make sure it still works.

n/a

[Yes](https://github.com/5e-bits/5e-srd-api/pull/330)

random photo - update me

### DIFF
--- a/src/5e-SRD-Features.json
+++ b/src/5e-SRD-Features.json
@@ -6937,7 +6937,7 @@
     "url": "/api/features/pact-magic"
   },
   {
-    "index": "eldritch-invocations-1",
+    "index": "eldritch-invocations",
     "class": {
       "index": "warlock",
       "name": "Warlock",
@@ -6953,7 +6953,7 @@
     ],
     "feature_specific": {
       "subfeature_options": {
-        "choose": 2,
+        "choose": "invocations known",
         "type": "feature",
         "from": {
           "option_set_type": "options_array",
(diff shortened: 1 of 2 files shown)


bagelbits (approved):
I think that makes sense. My only concern is that it makes it seem like all invocations are available. However, if you drill down, the features have a level prereq.

src/5e-SRD-Features.json
```diff
@@ -6953,7 +6953,7 @@
     ],
     "feature_specific": {
       "subfeature_options": {
-        "choose": 2,
+        "choose": "invocations known",
```
bagelbits: Actually. Thinking about this more. I have a concern with this now being either an int or a string. Why would we want to do this?

bagelbits: What does this get us?

SleeplessOne1917: I don't really know what other value to use in this instance. The number of invocations available to the warlock vary by level. While I'd certainly prefer to use an int, I don't exactly see how I can have choose equal the number of invocations known as an int as any one value I use will be incorrect.

bagelbits: I'm not sure if this change is a good idea in it's current state. I think this introduces inconsistency in the choice object that we were trying to get rid of with the Choice Refactor.

I wonder if a different way to approach this problem but have a similar outcome, is a way to extract out the list of Invocations into a list. That way you don't have to repeat it for every type? Then you can still have the choice for each level and the different number of invocations known. We're essentially doing the same thing as Known Spells, except it's a list of sub features instead of spells.

@fergcb?

SleeplessOne1917: Keep in mind that some invocations have levels in their prerequisites.

bagelbits: That is true. But prereqs are already set in each invocation. So that kind of handles itself?

SleeplessOne1917: I may not have followed your suggestion then. Nevermind.

Regardless, I'll ping @fergcb after work to ask his opinion on the PR.

fergcb: I think using a string in the `choose` field isn't a great solution. That said, I can't think of a solution that would be both intuitive and good for deserialisation.

Until we think of such a solution, my vote would just be to leave it as a list of all the invocations, with their prerequisites, and leave it to clients to implement the functionality.

SleeplessOne1917: > Until we think of such a solution, my vote would just be to leave it as a list of all the invocations, with their prerequisites,

@fergcb By this do you mean ditching the choice structure entirely and just have it be a list of the invocations? Perhaps giving the property on `feature_specific` its own name, like `invocations`?

fergcb: > By this do you mean ditching the choice structure entirely and just have it be a list of the invocations? Perhaps giving the property on `feature_specific` its own name, like `invocations`?

Yeah, exactly that.

SleeplessOne1917: I've done the suggested change.

bagelbits: So on the one hand, we have gotten rid of the string. However, on the other hand, we've gotten rid of the choice completely as well as the number of invocations you can choose at each level, which is not ideal either as we have now lost data.

I have an idea. Invocations are, essentially, like picking spells. Would it make sense to add a `invocations_known` field in the `class_specific` field of Levels? I can't think of a better place for this to live. This would both accomplish what @abias1122 is trying to resolve by cleaning up Features, but also we keep all of the current data portrayed.

I think this is also a breaking change as well.

@fergcb, thoughts?

fergcb: Yeah, sorry I should have clarified more when I suggested we swap the Choice for just a list of invocations. I had indeed figured that we would also be adding an `invocations_known` attribute somewhere.

Putting it in the `class_specific` object makes the most sense, I think. The only other alternative, I suppose, would be to have it as a single map in the Eldritch Invocations feature itself, but I don't like that nearly as much.

SleeplessOne1917: @bagelbits @fergcb `invocations_known` already exists on level class specific, so we should'nt need to make any more changes.

invocations known

SleeplessOne1917: Here's a bit of the data. Notice that there are 20 documents with `invocations_known`: one document for each of the 20 warlock levels. This is all in main btw.

invocations known search

bagelbits: Oh sweet! Thanks for pointing that out!

bagelbits (approved):

bagelbits:
This is a breaking change, no?

SleeplessOne1917:
> This is a breaking change, no?

Yes.

bagelbits:
Okay. I'll have to schedule this one then.

bagelbits:
This and the linked API change will get merged in on October 6th, 2022.

PR #500, opened 2022-09-02
feat: (src/5e-SRD-Monsters.json): added modifiers to hit dice

Stuart-Yee:
## What does this do?

Updates the hit dice stats for each of the monsters in src/5e-SRD-Monsters.json to have its proper die code modifier.
i.e. Hit Dice for Orc has been updated from 2d8 to 2d8+6

## How was it tested?

Spot checked several monsters against the SRD with at least one monster with a positive modifier (Orc), no modifier (Goblin) and a negative modifier (Kobold)

No

No

random photo - update me


fergcb:
I'm wondering whether the hit dice field should include the constitution modifier? `2d8` is the number and size of the orc's hit dice, whereas `2d8 + 6` is the calculation for rolling an orc's total HP.

These figures have different meanings, so maybe think they should be named differently?

The monster's hit dice can be derived from this figure by simply removing the modifier, but perhaps the name "hit_dice" is still a tad misleading? Equally, the modifier is simply the con mod × number of hit dice, so can already be calculated using API data.

This is pure pedantry, but I thought I'd raise the point just in case others agree that this could be a confusing change, which isn't really fixing anything.

Stuart-Yee:
> I'm wondering whether the hit dice field should include the constitution modifier? `2d8` is the number and size of the orc's hit dice, whereas `2d8 + 6` is the calculation for rolling an orc's total HP.
> 
> These figures have different meanings, so maybe think they should be named differently?
> 
> The monster's hit dice can be derived from this figure by simply removing the modifier, but perhaps the name "hit_dice" is still a tad misleading? Equally, the modifier is simply the con mod × number of hit dice, so can already be calculated using API data.
> 
> This is pure pedantry, but I thought I'd raise the point just in case others agree that this could be a confusing change, which isn't really fixing anything.

I see your point. My thoughts were just to make it match the SRD as that's how "Hit Dice" are listed there. I also didn't like have to calculate the mod on the client side, but maybe that's just my preference. I could simply add the hit dice with the calculated mod as a new field ("hit_dice_mod", or "roll_hit_points"?)

fergcb:
Yeah, I like that solution. In addition to keeping the two fields separate for their different uses, it'll also be less disruptive for existing consumers of the API.

Any of those suggested names should work. I would also throw `hit_points_roll` into the ring, as `roll_hit_points` sounds more like a verb/function name.

bagelbits:
I would like to second `hit_points_roll`

Stuart-Yee:
> I would like to second `hit_points_roll`

Cool. I'll probably make the change and commit it tomorrow.

Stuart-Yee:
> I would like to second `hit_points_roll`

Okay, committed changes. This should be ready to merge unless I should change anything else.

bagelbits (commented):
I see this in a few places, but it looks like we're replacing `’` with `\u00e2\u20ac\u2122` and `×` with `\u00c3\u2014`

I'm not sure if this is intentional, but we should probably undo that?

SleeplessOne1917:
I saw that this change would need a change on the API, so I went ahead and did that [here](https://github.com/5e-bits/5e-srd-api/pull/333).

bagelbits:
@Stuart-Yee Just wanted to make sure you saw my comment on the unicode characters.

Stuart-Yee:
> @Stuart-Yee Just wanted to make sure you saw my comment on the unicode characters.

I did, sorry I've been real slammed at work. I'll try to loop back to this by the end of the week if that's okay. The unicode changes were unintentional so I'll have to trace where that happened.

Stuart-Yee:
@bagelbits 
TD;LR version: I found the erroneous unicode characters, manually replaced them with the correct characters and pushed the changes.

The longer story: Originally I used a Python script to calculate the what the Hit Dice + CON modifier would be. The tricky bit was reassembling the JSON so that the `hit_points_roll` attribute would be inserted after the `hit_dice` attribute as opposed to simply appending the attribute to the end of each monster object. Clearly something in the script had a bug that changed those characters as it reassembled the JSON. I never did root trace the problem and since there were only a handful of occurrences, I simply executed a find and replace. If you're curious, I can share the script I used.

bagelbits:
I'm vaguely curious but not enough to debug right now. But if you through it in #help in the Discord, I can poke around when I have some free time. I know Python can be weird with strings some times.

I'll take a look a little later today!

Stuart-Yee:
> I'm vaguely curious but not enough to debug right now. But if you through it in #help in the Discord, I can poke around when I have some free time. I know Python can be weird with strings some times.
> 
> I'll take a look a little later today!

Sounds good, I'll post when I get a moment. 

It's a single-use script so I didn't write it very elegantly or for readability-- don't judge too harshly! 😂

bagelbits:
> It's a single-use script so I didn't write it very elegantly or for readability-- don't judge too harshly! 😂

Haha! No worries. I follow the methodology of:
Step 1: Make it work
Step 2: Make it pretty

bagelbits (approved):

PR #501, opened 2022-09-04
fix(wand-of-wonder): Fixed wand of wonder description table.

Aryathel:
## What does this do?

Changes the description for the `Wand of Wonder` Magic Item.

## How was it tested?

Was not tested, is a simple text update.

Not that I know of.

Shouldn't be necessary.


bagelbits (approved):
Thanks!

PR #509, opened 2022-11-01
Add some paladin spells

dadtmt:
## What does this do?

Adding following spells for Paladin:

- Compelled Duel
- Searing Smite
- Thunderous Smite
- Wrathful Smite


fergcb:
Unfortunately, none of these spells are in the SRD, and therefore we cannot legally distribute them, according to the terms of the license under which they are published.

You can read more about the SRD here: https://dnd.wizards.com/resources/systems-reference-document

PR #297, opened 2020-10-11
Add Monster Armor Types

txtsd:
## What does this do?
This adds monster armor types.

## How was it tested?
No tests. It was manually entered.

## Is there a Github issue this is resolving?
Yes #280 

## Did you update the docs in the API? Please link an associated PR if applicable.
No

## More
So right now it's a straight rip from the PDF. Let me know how you want the data actually formatted.
Here are the unique entries:
```
{'10 in humanoid form, 11 (natural armor) in bear and hybrid form',
 '10 in humanoid form, 11 (natural armor) in boar or hybrid form',
 '11 in humanoid form, 12 (natural armor) in wolf or hybrid form',
 '14 (natural armor), 11 while prone',
 '15 with _mage armor_',
 '16 with _barkskin_',
 'armor scraps',
 'barding scraps',
 'breastplate',
 'chain mail',
 'chain mail, shield',
 'chain shirt',
 'chain shirt, shield',
 'hide armor',
 'hide armor, shield',
 'leather armor',
 'leather armor, shield',
 'natural armor',
 'natural armor, shield',
 'patchwork armor',
 'plate',
 'scale mail',
 'scale mail, shield',
 'splint',
 'studded leather',
 'studded leather, shield'}
```
Words surrounded by underscores are in italics.

PR #515, opened 2022-11-19
feat: Add armor description to monsters

Wyrmwood:
## What does this do?

Adds a missing field from the SRD

## How was it tested?

Just loading and accessing, and manually verifying some of the new fields. Also spot checked apostrophe characters. 

No

No


bagelbits (commented):
Can you redo this without all the whitespace changes? What are the whitespace changes?

Also, you're going to have to do a PR in the [API](https://github.com/5e-bits/5e-srd-api) to updated docs, graphQL, and Typescript.

bagelbits:
The last two commits in the API should show you most of what you'll have to update (along with a bunch of other changes). Mostly its the graphql, types.d, and docs file in those two commits to look at.

bagelbits:
Thinking about this more. There is a bit more nuance on this that came up in a similar attempt in #297. I'm double checking what the data shape should be but once that's in place, we can go from there. The main issue is that a number of monsters have multiple ACs depending on conditions and spells.

bagelbits:
Closing in favor of https://github.com/5e-bits/5e-database/pull/518

PR #524, opened 2023-01-18
fix(monsters.json): updating to make the count property datatype consistent.

jfmcdonald:
## What does this do?

take 2 on the right repo this time.
This change updates the Monsters.json source file to set the multiattack count properties to a consistent datatype. In 99% of cases the count is an integer. There are however 2(?) instances where the count is a string. This is a pain when parsing with statically typed languages and brakes unmarshaling of the data. I couldn't think of a good way to convert the hydra for instance to a consistent integer so I went and retyped the property throughout.  I'm open to better ideas. 

## How was it tested?

lots of grep statements.

No, I probably should have created an issue, Sorry.

No, I don't think the datatypes are documented anywhere?

random photo - update me

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -103,7 +103,7 @@
         "actions": [
           {
             "action_name": "Tentacle",
-            "count": 3,
+            "count": "3",
             "type": "melee"
           }
         ]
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
Good catch. Violet fungus and Hydra kind of mess everything up.

I think we'll also need to update the types in the API as well.

bagelbits:
The data types are documented in a few different places in https://github.com/5e-bits/5e-srd-api

bagelbits:
I can own updating the API side.

bagelbits:
Closing as this was moved to the mono repo and merged

PR #526, opened 2023-01-28
fix: Add usage dict to recharge 6 actions

ASTRELION:
## What does this do?

This adds the following "usage" dict to the 7 actions with (Recharge 6) in their name and removes (Recharge 6) from the action name. I'm not sure if this was left out on purpose, or maybe I'm misunderstanding the Recharge rule, so feel free to close this if it's not applicable.
```json
"usage": {
  "type": "recharge on roll",
  "dice": "1d6",
  "min_value": 6
}
```

This may be a breaking change?

## How was it tested?

N/A

No

N/A

random photo - update me

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -6336,8 +6336,13 @@
         ]
       },
       {
-        "name": "Acid Spray (Recharge 6)",
+        "name": "Acid Spray",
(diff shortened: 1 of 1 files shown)


bagelbits (approved):
It's also on a short or long rest but yeah that is the correct usage of this. Good catch!

PR #534, opened 2023-05-01
feat: Add multiple missing spells

nicknanos:
## What does this do?

Add missing spells:
-Friends 
-Toll the Dead 
-Blade Ward  
-Dissonant Whisper 
-Cloud of Dagger 
-Crown of Madness 

No

No

### DIFF
--- a/src/5e-SRD-Spells.json
+++ b/src/5e-SRD-Spells.json
@@ -15815,5 +15815,254 @@
       }
     ],
     "url": "/api/spells/zone-of-truth"
+  },
+  {
+    "index": "friends",
+    "name": "Friends",
+    "desc": [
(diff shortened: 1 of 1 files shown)


fergcb:
Unfortunately, none of these spells are in the [D&D 5e SRD](https://dnd.wizards.com/resources/systems-reference-document), and therefore we can't legally distribute this content as part of the API.

You're of course welcome to maintain a private fork of the API for personal use containing any additional content that you own, but we won't be able to merge this PR, I'm afraid. Sorry to be the bearer of bad news!

strickinato:
I was surprised Crown of Madness was missing.
I found this issue and poked around that SRD link -- figured I'd slap this relevant bit from the FAQ here:

> **Why is the SRD missing some spells, magic items, and monsters?**
> In general, the criteria for what went into the SRD is if it (1) was in the 3E SRD, (2) has an equivalent in fifth edition D&D, and (3) is vital to how a class, magic item, or monster works. For example, the third edition SRD has the Delay Poison spell, but in fifth edition, that’s handled by the [Protection from Poison](https://www.dndbeyond.com/spells/2222-protection-from-poison) spell, so Protection from Poison is in the SRD.

PR #539, opened 2023-07-20
fix: Change Eldritch Blast damage to match Magic Missile

ASTRELION:
## What does this do?

Changes Eldritch Blast's damage to match the format of Magic Missile. I.e. Magic Missile lists damage rolls as *per missile*, even though you may cast multiple missiles. Eldritch Blast can also focus multiple targets or one like Magic Missile can, and the amount of blasts scale like Magic Missile does, so its damage should follow the same format. So, Eldritch Blast damage rolls are now *per beam*.

I'm not sure if there are more spells that operate similarly. Either way, I think they should all follow the same format.

I'd also like to informally suggest a new field for spells like this that defines how many times a spell is used per cast. So, for example, Eldritch Blast would have a field defining 1 beam at level 1, 2 beams at level 5, etc. I'm not sure if that's realistic, how many spells that impacts, or if it would be useful. I can open an actual issue for tracking that if it actually sounds useful to people.

## How was it tested?

Built locally.

No.

N/A

random photo - update me


bagelbits (approved):
Good catch. I think you should open an issue for casts at character level to better capture this .

ASTRELION:
I opened issues for the new field suggested
Database issue: https://github.com/5e-bits/5e-database/issues/541
API issue: https://github.com/5e-bits/5e-srd-api/issues/361

PR #559, opened 2023-12-08
feat: added additional feats that fit the current schema

TimoWess:
Simply adds a few aditional feats


fergcb (commented):
Unfortunately, these feats aren't printed in the SRD, and therefore we can't include them in the API. You're welcome to use them in your own private instance of the database, we just can't accept these changes into the public repo.

PR #561, opened 2024-01-13
fix: Normalize resistance/immunity/vulnerability strings

ASTRELION:
## What does this do?

Some of these strings would contain "...damage from nonmagical..." or "...from nonmagical...", so this PR removes "damage" from every string to be consistent with the rest.

Also, the Imp had a completely different format of "nonmagical/nonsilver" instead of the usual "nonmagical not made with silver".

Finally, I made the Archmage's resistance from Stoneskin more like the rest of the resistances' formats.

Some strings have "nonmagical attacks not made with silvered weapons" and some have "nonmagical weapons that aren't silvered" and those are unchanged. I wasn't sure if there was a practical difference between them, but if there isn't, I'd be happy to normalize those, too.

Let me know if I missed any!

*Force push was to correct misspelled commit message.*

## How was it tested?

Built locally.

No.

N/A

random photo - update me

### DIFF
--- a/src/5e-SRD-Monsters.json
+++ b/src/5e-SRD-Monsters.json
@@ -6705,7 +6705,7 @@
     "damage_vulnerabilities": [],
     "damage_resistances": [
       "damage from spells",
-      "nonmagical bludgeoning, piercing, and slashing (from stoneskin)"
+      "bludgeoning, piercing, and slashing from nonmagical attacks (from stoneskin)"
     ],
     "damage_immunities": [],
     "condition_immunities": [],
@@ -21385,7 +21385,7 @@
     "proficiencies": [],
     "damage_vulnerabilities": [],
     "damage_resistances": [
-      "bludgeoning, piercing, and slashing damage from nonmagical weapons"
+      "bludgeoning, piercing, and slashing from nonmagical weapons"
     ],
     "damage_immunities": [],
     "condition_immunities": [],
@@ -24089,7 +24089,7 @@
     "damage_vulnerabilities": [],
     "damage_resistances": [
       "cold",
-      "bludgeoning, piercing, and slashing from nonmagical/nonsilver weapons"
+      "bludgeoning, piercing, and slashing from nonmagical attacks not made with silvered weapons"
     ],
     "damage_immunities": [
       "fire",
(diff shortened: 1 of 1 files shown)


bagelbits:
@ASTRELION Can you list the additional strings that you're not sure of and how you might group them? As it stands, thanks for doing this, I think it was necessary.

src/5e-SRD-Monsters.json
```diff
@@ -24089,7 +24089,7 @@
     "damage_vulnerabilities": [],
     "damage_resistances": [
       "cold",
-      "bludgeoning, piercing, and slashing from nonmagical/nonsilver weapons"
+      "bludgeoning, piercing, and slashing from nonmagical attacks not made with silvered weapons"
```
bagelbits, line 24092: I think this is functionally the same as the ones below.

bagelbits (approved):

ASTRELION:
> @ASTRELION Can you list the additional strings that you're not sure of and how you might group them? As it stands, thanks for doing this, I think it was necessary.

For example, there are 4 instances of something like:
`"bludgeoning, piercing, and slashing from nonmagical attacks not made with silvered weapons"`

And 23 instances of something like:
`"bludgeoning, piercing, and slashing from nonmagical weapons that aren't silvered"`

Looking at the SRD, it does seem like they actually use those different strings. So, maybe it should just stay that way in the API.

bagelbits:
@ASTRELION I think that they're functionally the same so I don't mind rewriting them for consistency.

@fergcb What do you think?

fergcb:
> @fergcb What do you think?

Like you say, they're functionally the same. I'm in favour of this change - it'll make things nicer for computers, and it won't change the meaning for humans.

ASTRELION:
Are they actually the same? Couldn't it be interpreted that "bludgeoning, piercing, and slashing from nonmagical attacks not made with silvered weapons" would be resistant to an unarmed, bludgeoning attack but, "bludgeoning, piercing, and slashing from nonmagical weapons that aren't silvered" would not be resistant to an unarmed, bludgeoning attack since unarmed isn't a weapon. Or does an unarmed attack count as a weapon? Maybe there are better examples, but this was the first I thought of.

bagelbits:
@ASTRELION As far as I am aware, they are the same. An unarmed attack is a melee weapon attack.

ASTRELION:
Sorry for the delay, this should be ready to merge now.

The last commit changes all 4 of the outlying `"nonmagical attacks not made with silvered weapons"` strings to `"nonmagical weapons that aren't silvered"`.

bagelbits (approved):

Issue #790, opened 2025-03-03
natural-explorer-1-terrain-type missing choices


Dreaming-Codes:
natural-explorer-1-terrain-type is missing choices features

bagelbits:
You're right. It is missing choices for the terrains.

Dreaming-Codes:
How am I supposed to define parent, cause each terrain type should have multiple parents, natural-explorer-[1-3]-terrain-type.

bagelbits:
@Dreaming-Codes I would follow the pattern for `bard-expertise-[1-2]` Since you'll want to make a feature specific choice and each of the features is a choose one.

Dreaming-Codes:
> @Dreaming-Codes I would follow the pattern for `bard-expertise-[1-2]` Since you'll want to make a feature specific choice and each of the features is a choose one.

Not what I meant. What I was asking is what I should put in the parent field like here:

```json
{
  "index": "natural-explorer-arctic",
  "class": {
    "index": "ranger",
    "name": "Ranger",
    "url": "/api/2014/classes/ranger"
  },
  "name": "Natural Explorer: Arctic",
  "level": 6,
  "prerequisites": [],
  "desc": [
    "You have honed your survival skills in frozen, icy regions."
  ],
  "parent": {
    "index": "natural-explorer-2-terrain-types",
    "name": "Natural Explorer (2 terrain types)",
    "url": "/api/2014/features/natural-explorer-2-terrain-types"
  },
  "url": "/api/2014/features/natural-explorer-arctic"
}
```

How can I put multiple parents?

Dreaming-Codes:
I think https://github.com/5e-bits/5e-srd-api/blob/9c275e9040e7cf7bf3ea6488f370743c8906959b/src/models/2014/feature/index.ts#L42 should become `[APIReferenceSchema]`

fergcb:
@Dreaming-Codes I don't think creating sub-features in this way is the correct way to go about this. The player isn't actually presented with a choice of features at this level, but with a choice within this specific feature.

As @bagelbits mentions, it would make the most sense (from a consistency perspective, both with existing features and with the way the data is presented in the SRD), to follow the example of `bard-expertise-1/2`, and have a `feature_specific.terrain_type_options` field on these features, which allows the player to select a terrain type.

```json
{
  "index": "natural-explorer-1-terrain-type",
  "class": {
    "index": "ranger",
    "name": "Ranger",
    "url": "/api/2014/classes/ranger"
  },
[...]

PR #794, opened 2025-03-18
feat: add terrain type options to Natural Explorer features

Dreaming-Codes:
## Summary
- Added feature_specific field with terrain_type_options to natural-explorer-1-terrain-type, natural-explorer-2-terrain-types, and natural-explorer-3-terrain-types features
- Used string type in options as discussed in issue #790
- Fixed ranger natural explorer terrain type choices

Closes #790

PR #795, opened 2025-03-18
feat: add enemy type options to Favored Enemy features

Dreaming-Codes:
## Summary
- Added feature_specific field with enemy_type_options to favored-enemy-1-type, favored-enemy-2-types, and favored-enemy-3-enemies features
- Used string type in options as discussed in issue #790
- Fixed ranger favored enemy choice options

Related to #790


bagelbits (approved):
Similar to #794, once we get the API updated, these should be good to merge!

PR #824, opened 2025-04-28
Added all 2014 and 2024 feats

ToMakPo:
## What does this do?

I've added all of the 2014 and 2024 feats to the JSON files. Please note, the feats class was only setup to display ability score prerequisites. the 2014 feats also has spellcasting and armor training prerequisites, and 2024 has those as well as level prereqs. Because there was no plase to add those, I simply added them to the bottom of the description. I will create a feature request to add those also. Once it is ready, I can easily go back update the JSON files to include those.

## How was it tested?

I ran `npm run test` and all tests passed.

No

No

image

### DIFF
--- a/src/2014/5e-SRD-Feats.json
+++ b/src/2014/5e-SRD-Feats.json
@@ -1,5 +1,131 @@
 [
   {
+    "index": "actor",
+    "name": "Actor",
+    "prerequisites": [],
+    "desc": [
+      "Skilled at mimicry and dramatics, you gain the following benefits:",
+      "- Increase your Charisma score by 1, to a maximum of 20.",
+      "- You have advantage on Charisma (Deception) and Charisma (Performance) checks when trying to pass yourself off as a different person.",
+      "- You can mimic the speech of another person or the sounds made by other creatures. You must have heard the person speaking, or heard the creature make the sound, for at least 1 minute. A successful Wisdom (Insight) check contested by your Charisma (Deception) check allows a listener to determine that the effect is faked."
+    ],
+    "url": "/api/2014/feats/actor"
+  },
+  {
+    "index": "alert",
+    "name": "Alert",
+    "prerequisites": [],
+    "desc": [
+      "Always on the lookout for danger, you gain the following benefits:",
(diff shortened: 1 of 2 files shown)


src/2014/5e-SRD-Feats.json
```diff
@@ -1,4 +1,130 @@
 [
+  {
+    "index": "actor",
```
fergcb, line 3: Many of these feats are not in the SRD, so we cannot include them. Check our FAQ for more info about the SRD.

The only Feat in the 2014 SRD is "Grappler", and SRD 5.2 (2024) includes the origin feats Alert, Magic Initiate, Savage Attacker, and the general feat Grappler.

src/2024/5e-SRD-Feats.json
```diff
@@ -0,0 +1,1228 @@
+[
+  {
+    "index": "ability-score-improvement",
+    "name": "Ability Score Improvement",
+    "prerequisites": [],
+    "desc": [
+      "General Feat",
+      "Increase one ability score of your choice by 2, or increase two ability scores of your choice by 1. This feat can't increase an ability score above 20.",
+      "**Repeatable**. You can take this feat more than once.",
+      "PREREQ: level = 4"
```
fergcb, line 10: Same here. We should be using the `prerequisites` array.

src/2024/5e-SRD-Feats.json
```diff
@@ -0,0 +1,1228 @@
+[
+  {
+    "index": "ability-score-improvement",
+    "name": "Ability Score Improvement",
+    "prerequisites": [],
+    "desc": [
+      "General Feat",
```
fergcb, line 7: What purpose does this serve? The SRD doesn't contain this text in the Feat description.

We probably need to add a `feat_type` field to differentiate Origin, General, Fighting Style and Epic Boon Feats..

fergcb: I'm aware that you mention this in your PR description, but we should probably be designing this data nicely all in one go, rather than breaking it into lots of PRs.

PR #1014, opened 2026-03-02
feat(2024): Add Traits + Species

G-Ambatte:
## What does this do?

This PR adds the 2024 Traits and Species tables.

## How was it tested?

Local tests passed successfully.

Continued data entry from the 2024 SRD.

No

My friend is hugely into _Heated Rivalry_, so I made her this bookmark.
IceSports

### DIFF
--- a/src/2024/5e-SRD-Species.json
+++ b/src/2024/5e-SRD-Species.json
@@ -361,6 +361,23 @@
 				"url": "/api/2024/traits/otherworldly-presence",
 				"name": "Otherworldly Presence"
 			}
+		],
+		"subspecies": [
+			{
+				"index": "fiendish-legacy-abyssal",
+				"url" : "/api/2024/subspecies/fiendish-legacy-abyssal",
+				"name": "Fiendish Lineage: Abyssal"
+			},
+			{
+				"index": "fiendish-legacy-chthonic",
+				"url" : "/api/2024/subspecies/fiendish-legacy-chthonic",
+				"name": "Fiendish Lineage: Chthonic"
+			},
+			{
+				"index": "fiendish-legacy-infernal",
+				"url" : "/api/2024/subspecies/fiendish-legacy-infernal",
+				"name": "Fiendish Lineage: Infernal"
+			}
 		]
 	}
 ]
--- a/src/2024/5e-SRD-Species.json
+++ b/src/2024/5e-SRD-Species.json
@@ -79,7 +79,6 @@
 		"size": "Medium",
 		"speed": 30,
 		"traits": [
-
 			{
 				"index": "darkvision-120",
 				"url": "/api/2024/traits/darkvision-120",
--- a/src/2024/5e-SRD-Species.json
+++ b/src/2024/5e-SRD-Species.json
@@ -283,7 +283,24 @@
 		"url": "/api/2024/species/human",
 		"name": "Human",
 		"type": "Humanoid",
-		"size_options": ["Small", "Medium"],
+		"size_options": {
+			"desc": "Choose a size when you select this species.",
+			"choose": 1,
+			"type": "size_categories",
+			"from": {
+				"option_set_type": "options_array",
+				"options": [
+					{
+						"option_type": "size",
+						"size": "Small"
+					},
+					{
+						"option_type": "size",
+						"size": "Medium"
+					}
+				]
+			}
+		},
 		"speed": 30,
 		"traits": [
 			{
--- a/src/2024/5e-SRD-Species.json
+++ b/src/2024/5e-SRD-Species.json
@@ -198,7 +198,6 @@
 		"size": "Medium",
 		"speed": 35,
 		"traits": [
-
 			{
 				"index": "giant-ancestry",
 				"url": "/api/2024/traits/giant-ancestry",
--- a/src/2024/5e-SRD-Species.json
+++ b/src/2024/5e-SRD-Species.json
@@ -228,9 +228,9 @@
 				"name": "Adrenaline Rush"
 			},
 			{
-				"index": "darkvision",
-				"url": "/api/2024/traits/darkvision",
-				"name": "Darkvision"
+				"index": "darkvision-120",
+				"url": "/api/2024/traits/darkvision-120",
+				"name": "Darkvision (120 ft.)"
 			},
 			{
 				"index": "relentless-endurance",
@@ -251,7 +251,7 @@
 			{
 				"index": "darkvision-60",
 				"url": "/api/2024/traits/darkvision-60",
-				"name": "Darkvision (60ft.)"
+				"name": "Darkvision (60 ft.)"
 			},
 			{
 				"index": "fiendish-legacy",
--- a/src/2024/5e-SRD-Species.json
+++ b/src/2024/5e-SRD-Species.json
@@ -138,6 +138,23 @@
 				"url": "/api/2024/traits/trance",
 				"name": "Trance"
 			}
+		],
+		"subspecies": [
+			{
+				"index": "elven-lineage-drow",
+				"url" : "/api/2024/subspecies/elven-lineage-drow",
+				"name": "Elven Lineage: Drow"
+			},
+			{
+				"index": "elven-lineage-high-elf",
+				"url" : "/api/2024/subspecies/elven-lineage-high-elf",
+				"name": "Elven Lineage: High Elf"
+			},
+			{
+				"index": "elven-lineage-wood-elf",
+				"url" : "/api/2024/subspecies/elven-lineage-wood-elf",
+				"name": "Elven Lineage: Wood Elf"
+			}
 		]
 	},
 	{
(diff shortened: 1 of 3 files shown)


src/2024/5e-SRD-Species.json
```diff
@@ -0,0 +1,268 @@
+[
+	{
+		"index": "dragonborn",
+		"url": "/api/2024/species/dragonborn",
+		"name": "Dragonborn",
+		"type": "Humanoid",
+		"size": "Medium (about 5-7 feet tall)",
+		"speed": "30 feet",
```
bagelbits: Would it make sense for this to just be `30`? That would match races from 2014.

src/2024/5e-SRD-Species.json
```diff
@@ -0,0 +1,268 @@
+[
+	{
+		"index": "dragonborn",
+		"url": "/api/2024/species/dragonborn",
+		"name": "Dragonborn",
+		"type": "Humanoid",
```
bagelbits, line 6: [Out of scope]: I wonder whether we will want a type/subtype collection similar to `equipmentCategories`.

src/2024/5e-SRD-Species.json
```diff
@@ -0,0 +1,268 @@
...
+		"size": "Medium (about 5-7 feet tall)",
```
bagelbits: I would keep size consistent to Size categories (Small, Medium, etc). Additional data can be added to `size_description`, similar to races. This allows an easy query of `Give me all species of size Small`.

src/2024/5e-SRD-Species.json
```diff
@@ -0,0 +1,268 @@
...
+		"speed": "30 feet",
+		"description": "As a Dragonborn, you have these special traits.",
```
bagelbits: Do we really want this field? I would assume this would give more detail on the species.

src/2024/5e-SRD-Species.json
```diff
@@ -0,0 +1,268 @@
+[
+	{
+		"index": "dragonborn",
```
bagelbits, line 3: Are we missing a `subspecies` field on this and other `species`?

src/2024/5e-SRD-Species.json
```diff
@@ -0,0 +1,268 @@
...
+				"name": "Relentless Endurance"
+			}
+		]
+	},
+	{
+		"index": "tiefling",
+		"url": "/api/2024/species/tiefling",
+		"name": "Tiefling",
+		"type": "Humanoid",
+		"size": "Medium (about 4-7 feet tall) or Small (about 3-4 feet tall), chosen when you select this species",
```
bagelbits: Wow, I hate this. Should this become a `size_choice` or `size_options` field?

src/2024/5e-SRD-Subspecies.json
```diff
@@ -0,0 +1,487 @@
...
+				"traits" : [
+					{
+						"index": "darkvision-120",
+						"name": "Darkvision (120 ft.)",
+						"url": "/api/2024/traits/darkvision-120"
+					}
+				],
+				"spells": [
+					{ "nyi": "Dancing Lights" }
+				]
```
bagelbits: For the API, I'll just skip implementing this for now.

src/2024/5e-SRD-Subspecies.json
```diff
@@ -0,0 +1,487 @@
...
+	},
+	{
+		"index": "elven-lineage-wood-elf",
+		"url" : "/api/2024/subspecies/elven-lineage-wood-elf",
+		"name": "Elven Lineage: Wood Elf",
+		"effects": [
+			{
+				"level": 1,
+				"description": "Your Speed increases to 35 feet. You also know the Druidcraft cantrip.",
+				"traits": [],
```
bagelbits: Should this be here if it's empty? It seems like we only consistently show this on level 1. That seems confusing.

src/2024/5e-SRD-Traits.json
```diff
@@ -0,0 +1,176 @@
+[
+  {
```
bagelbits, line 2: Should we link the species/subspecies associated to each trait similar to 2014?

src/2024/5e-SRD-Traits.json
```diff
@@ -0,0 +1,176 @@
...
+    "index": "halfling-nimbleness",
+    "url": "/api/2024/traits/halfling-nimbleness",
+    "name": "Halfling Nimbleness",
+    "description": "You can move through the space of any creature that is a size larger than you, but you can't stop in the same space."
+  },
+  {
+    "index": "keen-senses",
+    "url": "/api/2024/traits/keen-senses",
+    "name": "Keen Senses",
+    "description": "You have proficiency in the Insight, Perception, or Survival skill."
```
bagelbits: Should we add `proficiency_choices` here?

src/2024/5e-SRD-Traits.json
```diff
@@ -0,0 +1,176 @@
...
+    "index": "resourceful",
+    "url": "/api/2024/traits/resourceful",
+    "name": "Resourceful",
+    "description": "You gain Heroic Inspiration whenever you finish a Long Rest."
+  },
+  {
+    "index": "skillful",
+    "url": "/api/2024/traits/skillful",
+    "name": "Skillful",
+    "description": "You gain proficiency in one skill of your choice."
```
bagelbits: Should we add `proficiency_choices` here?

src/2024/5e-SRD-Species.json
G-Ambatte: That makes sense, I'll apply that change now.

src/2024/5e-SRD-Species.json
G-Ambatte: All playable races in the SRDv5.2 are `Humanoid`, so at this point, the `type` field seems quite pointless - to quote Governor Marley: "When there's only one candidate, there's only one choice!"



From memory, Plasmoids and Autognomes are non-Humanoid players races from 2014, but they never actually appeared in the SRD.

---

I'll leave it as is for now.

src/2024/5e-SRD-Species.json
G-Ambatte: At this point, I've changed `size` to `size_options`, an array of strings of the valid sizes.

src/2024/5e-SRD-Species.json
G-Ambatte: At this point, I've removed the additional size description text, so it's either `Small` or `Medium` - those are the only two options for Player Characters in the SRD.

src/2024/5e-SRD-Subspecies.json
G-Ambatte: I'm backfilling these arrays with the non-trait traits, like `High Elf: Cantrip Versatility` or `Wood Elf: Movement Speed Increase`.

src/2024/5e-SRD-Traits.json
G-Ambatte: Added `proficiency_choices`.

src/2024/5e-SRD-Species.json
G-Ambatte: Should all be added now - Dragonborn, Elf, Gnome, Goliath, Tiefling all have `subspecies` arrays.

src/2024/5e-SRD-Species.json
G-Ambatte: Having checked all of them, it's consistently `As a X, you have the following traits:`. Seems pointless; I will remove them.

src/2024/5e-SRD-Subspecies.json
G-Ambatte: These should now all be filled or removed.

src/2024/5e-SRD-Traits.json
G-Ambatte: Added.

src/2024/5e-SRD-Traits.json
G-Ambatte: This should be added now.

src/2024/5e-SRD-Species.json
G-Ambatte: > [Out of scope]: I wonder whether we will want a type/subtype collection similar to `equipmentCategories`.

I suspect if a user were looking to customize the database, it might be useful to have an enum for `size` or `type` that could be extended (e.g. `enum size_options { SMALL, MEDIUM }` can be extended to `enum size_options { TINY, SMALL, MEDIUM, LARGE, HUGE, GARGANTUAN, [...] }` ).

Whether the potential value of this hypothetical situation warrants a shift in the actual effort currently being expended is another question entirely.

bagelbits: Oh I was thinking about monsters as well, but I see what you mean. That's not a good separation of concerns. An enum for type makes sense though. As well as size. I think it's okay as is. Enum would be an API level concern and just consistency on the DB level.

bagelbits (approved):

PR #1056, opened 2026-04-02
feat(2024): Add 2024 Features

G-Ambatte:
## What does this do?

This is an initial pass at adding Feature data using the new Zod testing ensuring data consistency.

It includes a slight modification to testing to allow URLs ending in `-nyi` to bypass testing as they are expected to not exist yet.
When this occurs, the file and URL are output to the console.

- [ ] Adding a note here for future review: check that the Druid Wild Shape and Circle of the Land subclass tables (and any others) are still correctly formatted.

## How was it tested?

Locally.

No

No

image


src/2024/en/5e-SRD-Features.json
```diff
@@ -0,0 +1,2811 @@
+[
+{
+	"index": "barbarian-rage",
+	"url": "/api/2024/features/barbarian-rage",
+	"name": "Rage",
+	"class": {
+		"index": "barbarian",
+		"name": "Barbarian",
+		"url": "/api/2024/classes/barbarian-nyi"
+	},
```
bagelbits, line 10: I believe all classes are in so we can drop nyi

src/2024/en/5e-SRD-Features.json
```diff
@@ -0,0 +1,2811 @@
...
+	"class": {
+		"index": "druid",
+		"name": "Druid",
+		"url": "/api/2024/classes/druid-nyi"
+	},
+	"subclass": {
+		"index": "circle-of-the-land",
+		"name": "Circle of the Land",
+		"url": "/api/2024/classes/circle-of-the-land-nyi"
+	},
```
bagelbits, line 1168: I think these should be `/api/2024/subclasses/*` also these should all exist and no longer need `*-nyi`.

src/2024/en/5e-SRD-Features.json
G-Ambatte: `classes/(.*)-nyi` is changed to `classes/$1` in tonight's commit.

src/2024/en/5e-SRD-Features.json
G-Ambatte: These are also corrected in tonight's commit.

bagelbits (approved):

PR #1074, opened 2026-04-17
feat(2024): monster schema and first monsters

scarlettcsung:
## What does this do?

1. the schema now includes "in_lair" for xp and special ability usage times, to reflect new SRD formatting
2. moved Spellcasting up so it can be used in Actions, where spellcasting usually sits now
3. added aboleth and adult black dragon stat blocks

## How was it tested?

npm test
the only failures are non-existent links for spells, but I don't really know what to do with that as I understand someone else is working on spells

kind of? people have been asking about 2024

nope

image


bagelbits (commented):
I'm really sorry. Can you move these into `src/2024/en`?

You'll also want to hook your schema into `src/2024/tests/schemas.test.ts` and `src/2024/tests/translations.test.ts`

bagelbits (approved):

PR #1122, opened 2026-05-13
feat: Add feat 2014 EN

Quenway:
## What does this do?

Implements feat 2014 in English
Update structure : 

- ability_score became an array, for multiple ability_score and gain an entry type which is OR|AND
- prerequisites > race, feat, level, class added from scratch but with the same structure
- source : new entry also, in kebab-case, which is [see image]

## How was it tested?

Locally

nope

nope

No troubles here :)


bagelbits (commented):
Unfortunately, none of these feats are in the SRD.

PR #1152, opened 2026-06-03
feat: Add Docker argument for MongoDB version

### LINKED ISSUE #605: Publish multi-arch (arm64) Docker image
So, living in a system admin world all my adult / professional life I get the need to keep systems up to date and software up to date.  However I have to ask:  Is there a solution to MongoDB 5 requiring the AVX instruction set?  I realize that AVX was shipped with Haswell back in 2013 and thus 11 years is more than enough in the technical world to assume most people are running on it, but Xeon processors are RAMPANT in many home labs given the costs of systems.

Would it be too much to ask to release a dockerfile version for a Pre 5.x MongoDB installation that would work on older server hardware?  It's not as if 5eAPI requires a honking system to run, the restriction is simply in the chosen software version at build time.

G-Ambatte:
## What does this do?

This PR modifies the Dockerfile to allow it to take a parameter to select the MongoDB version that the image will be built from.

## How was it tested?

Locally using Docker Desktop

This resolves #605.

The API is unaffected by this change; however I did update the README.md to describe the modified build command.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,9 @@
 # This recipe is pulled from: https://stackoverflow.com/a/33601894
 
 # Parent Dockerfile https://github.com/docker-library/mongo/blob/982328582c74dd2f0a9c8c77b84006f291f974c3/3.0/Dockerfile
-FROM mongo:6.0.1
+ARG MONGODB='6.0.1'
+
+FROM mongo:${MONGODB}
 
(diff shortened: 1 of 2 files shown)


bagelbits:
I don't think this actually gives us anything, as you'd still need either:
* rebuild locally, which is the current workaround
*  have the CI build this and the normal version, but that might cause the API to break for the older version of mongo

